PR #11, opened 2022-06-03
Make sfu-to-sfu talk MSC3401 and add new features

ara4n:
Fixes https://github.com/matrix-org/sfu-to-sfu/issues/13
Implements [MSC3401](https://github.com/matrix-org/matrix-spec-proposals/pull/) SFU bits
Requires https://github.com/matrix-org/matrix-js-sdk/pull/2423
Requires https://github.com/mautrix/go/pull/77


src/conference.go
```diff
@@ -0,0 +1,226 @@
...
+type LocalTrackInfo struct {
+	StreamID string
+	TrackID  string
+	Call     *Call
+}
+
+type LocalTrackWithInfo struct {
+	Track *webrtc.TrackLocalStaticRTP
+	Info  LocalTrackInfo
+}
```
SimonBrandner, line 38: This now looks like an idea much worse than when I first thought of it... This could definitely be done differently now though I think it makes sense to just try to redo this during switching to using a pool of streams since that might change the design again...

src/call.go
```diff
@@ -0,0 +1,447 @@
...
+	CallID                 string
+	UserID                 id.UserID
+	DeviceID               id.DeviceID
+	LocalSessionID         id.SessionID
+	RemoteSessionID        id.SessionID
+	Client                 *mautrix.Client
+	PeerConnection         *webrtc.PeerConnection
+	Conf                   *Conference
+	DataChannel            *webrtc.DataChannel
+	LastKeepAliveTimestamp time.Time
```
dbkr, line 41: Do all of these need to be public? eg. LastKeepAliveTimestamp doesn't look like its used outside of the class.

src/conference.go
```diff
@@ -0,0 +1,240 @@
...
+	for streamID, info := range c.Metadata.Metadata {
+		_, exists := metadata[streamID]
+		if info.DeviceID == deviceID && !exists {
+			delete(c.Metadata.Metadata, streamID)
+		}
+	}
+}
+
+func (c *Conference) GetRemoteMetadataForDevice(deviceID id.DeviceID) event.CallSDPStreamMetadata {
+	metadata := make(event.CallSDPStreamMetadata)
```
dbkr, line 198: This function could use a comment or two

src/focus.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+type Focus struct {
+	name  string
+	confs Confs
+}
+
+func (f *Focus) Init(name string) {
+	f.name = name
+	f.confs.confs = make(map[string]*Conference)
```
dbkr, line 38: Generally this would be done in a constructor-like NewFocus function

src/main.go
```diff
@@ -0,0 +1,143 @@
...
+		log.Fatalf("failed to read config: %s", err)
+	}
+	var config Config
+	if err := yaml.Unmarshal(file, &config); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal YAML: %s", err)
+	}
+	return &config, nil
+}
+
+func OnKill(c chan os.Signal, beforeExit []func()) {
```
dbkr, line 108: `killListener` maybe, to capture the fact that it's waiting for a signal rather than a handler as such? (also don't think this needs to be public?)

src/matrix.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		if call, err = conf.GetCall(callID, false); err != nil || call == nil {
+			log.Printf("failed to get call %s: %s", callID, err)
+			return nil, err
+		}
+		return call, nil
+	}
+
+	syncer.OnSync(func(resp *mautrix.RespSync, since string) bool {
+		for _, evt := range resp.ToDevice.Events {
```
dbkr, line 70: Is this really the only way to get todevice events in mautrix-go?

dbkr (commented):
Generally looking good. Main things are proper (or at least some) error handling rather than panicking and making sure member functions aren't public unnecessarily. We'll also want to hook up logrus or whatever people use now at some point soon, but probably not in this PR.

scripts/run.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+clear && go run ./src/*.go --logTime
```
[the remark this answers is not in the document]
SimonBrandner: It correctly assumes Šimon wants his terminal content cleared though :P

src/call.go
```diff
@@ -0,0 +1,447 @@
...
+
+		default:
+			log.Fatalf("Unknown operation %s", msg.Op)
+			// TODO: hook up msg.Stop to unsubscribe from tracks
+		}
+	})
+}
+
+func (c *Call) NegotiationNeededHandler() {
+	offer, err := c.PeerConnection.CreateOffer(nil)
```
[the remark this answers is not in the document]
SimonBrandner: We log sending the event, so it seems redundant unless you want to log something specific?

src/matrix.go
SimonBrandner: Not sure, this was originally written by M - /me goes ask it the Maunium room

src/call.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	c.SendDataChannelMessage(event.SFUMessage{
+		Op:  event.SFUOperationOffer,
+		SDP: offer.SDP,
+	})
+}
+
+func (c *Call) IceCandidateHandler(candidate *webrtc.ICECandidate) {
+	if candidate == nil {
```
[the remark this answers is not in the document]
SimonBrandner: Hmm, how would the typing be solved in Go? `Candidates` can only ever `[]CallCandidate`

scripts/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+go build -o dist/bin src/*.go
```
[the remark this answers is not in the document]
SimonBrandner: To be perfectly honest, I am actually not sure what is the proper way to do this if the files are in a subdirectory...

SimonBrandner:
> We'll also want to hook up logrus or whatever people use now at some point soon, but probably not in this PR.

Yep, that has been my thoughts since the fist time I actually need to debug it based on logs :P

Sorry, for the Go newbie mistakes, hopefully, it looks better now

src/call.go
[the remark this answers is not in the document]
dbkr: oh, if that gets logged that's probably ok

src/call.go
```diff
@@ -235,6 +235,15 @@ func (c *Call) iceCandidateHandler(candidate *webrtc.ICECandidate) {
 
 	jsonCandidate := candidate.ToJSON()
 
+	callCandidate := event.CallCandidate{
+		Candidate:     jsonCandidate.Candidate,
+		SDPMLineIndex: int(*jsonCandidate.SDPMLineIndex),
+		SDPMID:        *jsonCandidate.SDPMid,
+	}
+	if jsonCandidate.UsernameFragment != nil {
+		callCandidate.UsernameFragment = *jsonCandidate.UsernameFragment
```
dbkr, line 244: also the password if we're doing this

dbkr (commented):
Looks good other than the ice password. Presumably we're blocked on the mautrix-go PR now?

scripts/build.sh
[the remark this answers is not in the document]
dbkr: ah yes - I think the 'expected' thing in Go is to have all the files at the top level, but personally I much prefer them all in a src directory. Some other projects just resort to having a makefile, so in that case I guess a script is fine.

src/call.go
SimonBrandner: There seems to be no way to do that? EW also doesn't seem to be sending it

src/call.go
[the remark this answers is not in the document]
dbkr: ah yes - we may need to store the fact that we've reached end-of-candidates separately, but it actually looks like we just listen for the gathering finished event anyway.

src/call.go
dbkr: Right yes, sorry - I got confused. The ice ufrag / password are just in the SDP rather than the candidates, so they don't need to be here at all. I'm not sure why it's on `IceCandidateInit`.

dbkr (approved):

PR #29, opened 2022-08-24
Refactor code

SimonBrandner:


src/call.go
```diff
@@ -114,12 +134,21 @@ func (c *Call) onDCUnpublish(stop []event.SFUTrackDescription, sdp string) {
...
+		for _, publisher := range c.Publishers {
+			if publisher.Matches(trackDesc) {
+				publisher.Stop()
+			} else {
+				newPublishers = append(newPublishers, publisher)
+			}
 		}
+		c.mutex.RUnlock()
+
+		c.mutex.Lock()
```
dbkr, line 149: probably better to just obtain a write mutex in the first place unless the stop / append operation is likely to take a long time?

src/call.go
```diff
@@ -541,3 +546,19 @@ func (c *Call) CheckKeepAliveTimestamp() {
...
+
+	c.mutex.RLock()
+	for _, subscriber := range c.Subscribers {
+		if subscriber != toDelete {
+			newSubscribers = append(newSubscribers, subscriber)
+		}
+	}
+	c.mutex.RUnlock()
+
+	c.mutex.Lock()
```
dbkr, line 561: same mutex comment

dbkr (approved):

PR #41, opened 2022-11-12
Containerize the SFU and publish images to the registry

daniel-abramov:
Now building an image is part of our CI pipeline for `main` and pull requests.

The package is published to the GitHub Container Registry so that we can easily access the image during the deployment pipeline.

Note that the deployment pipeline is missing here which makes the PR not that useful **right now**, but once we start deploying from a CD pipeline, this all starts making a lot of sense and would allow us to deploy as easily as `scp`ing a single `docker-compose.yml` to the server and then executing an Ansible action.

I also updated the way how to pass the parameters to the image (regular environment variable as it's the most common practice that is portable and compatible with k8s).

Resolves https://github.com/matrix-org/waterfall/issues/36.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,47 @@
+name: Build and publish Docker image
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request: {}
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Log in to the Container registry
-        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        uses: docker/login-action@v2
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
@@ -30,16 +30,16 @@ jobs:
 
       - name: Extract metadata (tags, labels) for Docker
         id: meta
-        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        uses: docker/metadata-action@v4
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=ref,event=branch
+            type=sha,format=short,event=branch
             type=ref,event=pr
             type=semver,pattern={{version}}
 
       - name: Build and push Docker image
-        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        uses: docker/build-push-action@v3
         with:
           context: .
           push: true
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,36 @@
+##
+## Build
+##
+FROM golang:1.19 AS build
+
+WORKDIR /app
+
+COPY go.mod ./
+COPY go.sum ./
+
+# Cache dependencies before building and copying the source code, so that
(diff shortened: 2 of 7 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
+version: "3.3"
+services:
+  waterfall:
+    container_name: waterfall
+    image: ghcr.io/matrix-org/waterfall:main
+    network_mode: host
+    restart: always
+    environment:
```
SimonBrandner, line 8: Is there a way to avoid using env vars? We've just started transitioning from using env vars to using a config file, so I think we'd prefer not to use env vars in general

README.md
```diff
@@ -44,3 +44,29 @@ from the MSC.
...
 * `./dist/bin`
 * Access at <http://localhost:8080>
+
+## Docker
+
+### Building
+
+In order to build a docker image, run the following in the root directory:
+
+`$ docker build . -t matrix/materwall`
```
SimonBrandner, line 54: Why `matrix/`? Perhaps I am missing some docker knowledge here?

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+COPY go.sum ./
+
+# Cache dependencies before building and copying the source code, so that
+# we don't need to re-download when building and so that the change in the
+# source code do not invalidate our downloaded layer.
+RUN go mod download
+
+COPY ./src ./src
+
+RUN go build -o /sfu ./src
```
SimonBrandner, line 18:
```suggestion
RUN go build -o /waterfall ./src
```

I think we'd prefer to call the binary `waterfall`

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,49 @@
+name: Build and public Docker image
```
SimonBrandner, line 1:
```suggestion
name: Build and publish Docker image
```

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+    branches:
+      - 'main'
```
SimonBrandner, line 9:
```suggestion
  pull_request: {}
```

I thought to run this for all PRs it would need to be this, am I wrong?

SimonBrandner (commented):
LGTM, just a few nits and a concern about env vars

docker-compose.yaml
daniel-abramov: Good question, let me elaborate on that 🙂 

So basically it all comes down to the flexibility of the setup so that we remain compatible with different ways to deploy and run the server as an image. Environment variables do not seem like the best idea when testing things locally or running the server in a bare Linux environment (which is the reason why we don't really obsolete the usage of the config file, it's still an option), however, it does not necessarily play well when we talk about automated deployments with e.g. k8s.

The problem with the config file is that it poses the question: how do we pass the config file to the image? Essentially we could do either of these:

- Embed the config file directly into the built image, i.e. making it part of the image. This is a bit problematic as it means that we must preconfigure the server at the moment the image is built which makes the image pretty much useless for any use case outside of the environment, the config of which was bundled together with the image.
- Another possibility is to supply the config file together with the image, but it means we would need to deliver an additional file with our `docker-compose` and mount the file to be accessible to the container on a runtime (while making sure that permissions, etc are all set right), in addition, we would need to pass the file name of the config explicitly as part of the `docker-compose.yml` which also makes the compose file a bit less useful.

This is normally solved by just passing the environment variables to the image as it also makes it easy to configure them per environment, i.e. we just deploy a single `docker-compose.yml` and we might provide a `.env` file that is automatically read when the container is started, meaning that the `docker-compose.yml` is environment-agnostic (could be used for both development and production environments) and all the configuration is just a part of the `.env` file as simple pairs of key=value.

There is an issue with the way I implemented which I realized upon reading your comment: since I introduced a separate variable for each configuration, it means that any time we add an additional configuration, we must both update the format of the config file **and** add handling of an additional environment variable to the function that reads environment variables. This is not very nice (copy-paste and syncing logic in two places), so I'm thinking about eliminating it the same way LiveKit SFU implemented it! What they did is that their `docker-compose.yml` file has a variable called `LIVEKIT_CONFIG` which is simply just a YAML. I.e. instead of creating separate variables for all configs, they allowed passing a `LIVEKIT_CONFIG` variable which is a regular body of the config file.

Perhaps I need to update it to have the same behavior to remain flexible and lower our maintenance effort!

README.md
daniel-abramov: It's just a tag, you can use any tag you want locally 🙂 I just scoped it for `matrix` to avoid confusion (in case there is another waterfall on a system 🤔) + it played well with the path to the `waterfall` in our container registry, so I thought for consistency it would be less confusing.

Dockerfile
daniel-abramov: Ok, let me change it to `waterfall` in all places. No strong opinion :)

.github/workflows/docker.yaml
daniel-abramov: I think you're right. What I wrote was meant to be "generate it for the pull requests that target a merge to the `main` branch". But I don't have any preference, we can do it the same way you did 🙂

docker-compose.yaml
daniel-abramov: @SimonBrandner, updated the PR to address all your points, ptal 🙂

.github/workflows/docker.yaml
daniel-abramov: Updated.

.github/workflows/docker.yaml
daniel-abramov: Done.

src/main.go
```diff
@@ -143,8 +116,14 @@ func main() {
...
-		logrus.WithError(err).Fatal("failed to load config file")
+
+	config, err = loadConfigFromEnv()
+	if err != nil {
+		logrus.WithError(err).Info("failed to load config from environment variables, trying to load from file")
+
+		if config, err = loadConfigFromPath(*configFilePath); err != nil {
+			logrus.WithError(err).Fatal("failed to load config file")
+		}
 	}
```
SimonBrandner, line 127: Could this please be wrapped in a `loadConfig` function located in `config.go`?

SimonBrandner (approved):
LGTM otherwise!

src/main.go
daniel-abramov: Sure! Done.

SimonBrandner (approved):

Issue #39, opened 2022-11-10
Define a clear architecture for the SFU and refactor the current implementation


SimonBrandner:
@daniel-abramov:

The main objective of this issue is to make sure that we're building all new development in this project on top of safe decoupled abstractions that could be developed and unit-tested separately from each other and allow us to gain more speed and flexibility while extending the server!

Since SFU is slowly gaining more features and must be written in a future-proof way, it makes sense to make sure that we're building the SFU on a stable basis:

- [x] Enable as many links as possible.
- [x] Remove circular dependencies (pointers) between structures.
- [x] Confine access of objects as much as possible to each other (least privilege principle).
- [x] Define clear responsibilities of different parts of the system.
- [x] Remove variables that are just copies of the information available through other sources to avoid sync issues and confusion.
- [x] Lower the probability of race conditions by checking the use of all mutexes and replacing some of them with message-passing workflows (i.e. the "Do not communicate by sharing memory; instead, share memory by communicating.”)
- [x] Extract signaling logging in a separate file or package.
- [x] Document the code so that it's clear for application developers to understand what's going on and why it's going on 🙂

PR #42, opened 2022-11-14
Introduce stricter lints

daniel-abramov:
This relates to https://github.com/matrix-org/waterfall/issues/39

This PR does not really refactor anything, but only deals with the lining (introduces more strict lints, and turns off linters that are officially deprecated). The refactoring will be done on top of that.

Primarily it deals with global variables that had to be removed to satisfy the linter (also, global mutable variables a.k.a. singletones are kind of dangerous things in all languages).

This PR is intended to be merged after https://github.com/matrix-org/waterfall/pull/41

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,20 +1,32 @@
 linters-settings:
-  lll:
-    line-length: 120
+  nlreturn:
+    block-size: 2
   wsl:
-    allow-cuddle-declarations: true
+    enforce-err-cuddling: true
+  funlen:
+    lines: 100
+    statements: 50
+  gomoddirectives:
+    replace-allow-list:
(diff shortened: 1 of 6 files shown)


.golangci.yaml
```diff
@@ -1,20 +1,32 @@
 linters-settings:
-  lll:
-    line-length: 120
```
SimonBrandner, line 3: Why did this get removed?

SimonBrandner (approved):
LGTM

.golangci.yaml
daniel-abramov: It's [a default setting](https://golangci-lint.run/usage/linters/#lll) for the `lll` plugin 🙂

SimonBrandner (approved):

Issue #50, opened 2022-11-17
Implement efficient structured logging


daniel-abramov:
We do use a framework that allows structured logging, however currently we use different formats to log stuff in other places. We must unify them in order to have structured Grafana-compatible logging.

- [x] Log important key=value pairs when receiving signaling messages or when things happen
- [ ] Use a format that would be readable and compatible with Grafana
- [ ] Replace `logrus` with a more efficient framework. Unfortunately `logrus` falls short [when it comes to performance](http://hackemist.com/logbench/) 🙁 
- [ ] Group common logs together to reduce the amount of lines that we log.

daniel-abramov:
FWIW: Also, I've noticed that with `logrus` log timestamps are not always going up. I.e. several go-routines logging via `logrus` may result in the following log entries in the file:

```go
18:35:00
18:35:01
18:35:02
18:35:01
18:35:03
18:35:02
```

Which is not very convenient.

daniel-abramov:
After investigating several logging frameworks and while working on https://github.com/matrix-org/waterfall/issues/141 it seems like the best option would be to use `zerolog` (it's faster and a bit more ergonomic than `zap`), although we would need to implement a hook for the event handling in order to produce OpenTelemetry span events when e.g. info logs are logged. For more info see updated description of https://github.com/matrix-org/waterfall/issues/141

speatzle:
> After investigating several logging frameworks and while working on #141 it seems like the best option would be to use `zerolog` (it's faster and a bit more ergonomic than `zap`)

Have you looked at slog?
slog is the new "standard" golang logging package which has been proposed to be added to the standard library.
One of its main goals is to be performant.

I have been using slog in my personal and company project's over the past couple of months and am very pleased with it.

You can find the design doc [here](https://go.googlesource.com/proposal/+/master/design/56345-structured-logging.md).

For now it is available in `golang.org/x/exp/slog` but since the proposal was [accepted](https://github.com/golang/go/issues/56345#issuecomment-1470506816) by rsc about a day ago it should be in the standard library for the next go release.

PR #52, opened 2022-11-23
SFU Refactoring

### LINKED ISSUE #39: Define a clear architecture for the SFU and refactor the current implementation

daniel-abramov:
This closes https://github.com/matrix-org/waterfall/issues/39 and does add structured logging, but not the framework for https://github.com/matrix-org/waterfall/issues/50

The main idea of this refactoring was to provide us with a somewhat stable foundation on top of which we could build new features with as much compile-time safety and with some guarantees that we lower race conditions. 

See https://github.com/matrix-org/waterfall/issues/39 for a description of things that were improved and fixed.

Summary: [TODO, write some better explanation and highlights of this refactoring].

A rough diagram of how components are conceptually connected with each other:
waterfall-refactored

NB: This is still work-in-progress and **must be tested** before merging.


dbkr (commented):
This is feeling like a nice refactor so far - using channels & messaging passing in favour of locks definitely feels like a step in the right direction and is good to get done while the project is still a fairly manageable size.

pkg/signaling/client.go
```diff
@@ -0,0 +1,67 @@
...
+	client.DeviceID = whoami.DeviceID
+
+	return &MatrixClient{
+		client: client,
+	}
+}
+
+// Starts the Matrix client and connects to the homeserver,
+// Returns only when the sync with Matrix fails.
+func (m *MatrixClient) RunSync(callback func(*event.Event)) {
```
SimonBrandner, line 38: As far as I understand it, this starts syncing in a loop, could the function name indicate that?

pkg/conference/start.go
```diff
@@ -0,0 +1,64 @@
...
+	"maunium.net/go/mautrix/id"
+)
+
+// Starts a new conference or fails and returns an error.
+func StartConference(
+	confID string,
+	config Config,
+	signaling signaling.MatrixSignaling,
+	conferenceEndNotifier ConferenceEndNotifier,
+	UserID id.UserID,
```
SimonBrandner, line 34:
```suggestion
	userID id.UserID,
```

pkg/conference/start.go
```diff
@@ -0,0 +1,64 @@
...
+		config:         config,
+		signaling:      signaling,
+		matrixMessages: receiver,
+		endNotifier:    conferenceEndNotifier,
+		participants:   make(map[ParticipantID]*Participant),
+		peerMessages:   make(chan common.Message[ParticipantID, peer.MessageContent], common.UnboundedChannelSize),
+		logger:         logrus.WithFields(logrus.Fields{"conf_id": confID}),
+	}
+
+	participantID := ParticipantID{UserID: UserID, DeviceID: inviteEvent.DeviceID, CallID: inviteEvent.CallID}
```
SimonBrandner, line 50: I think I'd prefer this to be done outside this method, it feels like it shouldn't be here

pkg/signaling/matrix.go
```diff
@@ -0,0 +1,148 @@
...
+
+const LocalSessionID = "sfu"
+
+// Matrix client scoped for a particular conference.
+type MatrixForConference struct {
+	client       *mautrix.Client
+	conferenceID string
+}
+
+// Create a new Matrix client that abstarcts outgoing Matrix messages from a given conference.
```
SimonBrandner, line 34:
```suggestion
// Create a new Matrix client that abstracts outgoing Matrix messages from a given conference.
```

pkg/conference/start.go
daniel-abramov: The creation of the `participantID`?

pkg/conference/processor.go
```diff
@@ -0,0 +1,220 @@
...
+		participant.streamMetadata = sfuMessage.Metadata
+
+		participant.sendDataChannelMessage(event.SFUMessage{
+			Op:       event.SFUOperationAnswer,
+			SDP:      answer.SDP,
+			Metadata: c.getAvailableStreamsFor(participant.id),
+		})
+
+	case event.SFUOperationUnpublish:
+		participant.logger.Info("Received unpublish over DC")
```
SimonBrandner, line 196: We should be resending the metadata here and removing the track, right?

pkg/conference/processor.go
```diff
@@ -0,0 +1,220 @@
...
+
+		// TODO: Clarify the semantics of unpublish.
+	case event.SFUOperationAlive:
+		// FIXME: Handle the heartbeat message here (updating the last timestamp etc).
```
SimonBrandner, line 200: This needs to be fixed before we merge this, I think

pkg/conference/processor.go
```diff
@@ -0,0 +1,220 @@
...
+	case event.SFUOperationMetadata:
+		participant.streamMetadata = sfuMessage.Metadata
+		c.resendMetadataToAllExcept(participant.id)
+	}
+}
+
+func (c *Conference) processMatrixMessage(msg MatrixMessage) {
+	switch ev := msg.Content.(type) {
+	case *event.CallInviteEventContent:
+		c.onNewParticipant(msg.Sender, ev)
```
SimonBrandner, line 210: Perhaps it should be named `onInvite()`?

pkg/peer/webrtc.go
```diff
@@ -0,0 +1,171 @@
...
+	p.sink.Send(RenegotiationRequired{Offer: &offer})
+}
+
+// A callback that is called once we receive an ICE connection state change for this peer connection.
+func (p *Peer[ID]) onICEConnectionStateChanged(state webrtc.ICEConnectionState) {
+	p.logger.WithField("state", state).Info("ICE connection state changed")
+
+	switch state {
+	case webrtc.ICEConnectionStateFailed, webrtc.ICEConnectionStateDisconnected:
+		// TODO: Ask Simon if we should do it here as in the previous implementation.
```
SimonBrandner, line 108: The previous implementation doesn't handle this afaik, so we don't have to worry about this for now I think

pkg/peer/webrtc.go
```diff
@@ -0,0 +1,171 @@
...
+		//       Ideally we want to perform an ICE restart here.
+		// p.notify <- PeerLeftTheCall{sender: p.data}
+	case webrtc.ICEConnectionStateCompleted, webrtc.ICEConnectionStateConnected:
+		// FIXME: Start keep-alive timer over the data channel to check the connecitons that hanged.
+		// p.notify <- PeerJoinedTheCall{sender: p.data}
```
SimonBrandner, line 113: Do we ever send an event to let the peer know about the end of candidates? I think I've seen it somewhere but it should perhaps be here?

pkg/conference/matrix.go
```diff
@@ -0,0 +1,124 @@
...
+type MessageContent interface{}
+
+type MatrixMessage struct {
+	Sender   ParticipantID
+	Content  MessageContent
+	RawEvent *event.Event
+}
+
+// New participant tries to join the conference.
+func (c *Conference) onNewParticipant(participantID ParticipantID, inviteEvent *event.CallInviteEventContent) error {
```
SimonBrandner, line 20: Perhaps `onInvite()` instead

SimonBrandner (commented):
This is largely looking good to me though I sometimes struggle a bit not to get lost... Perhaps a better file naming would be a bit, some of the file names aren't very descriptive from my POV. 

I am also a little concerned about completely removing the subscriber/publisher model which is very useful for simulcast and possibly other things; it's also very similar to what ion-sfu and livekit does, iirc

I would also like my PR to get merged before this and for us to somehow integrate it into this one

All of the TODO are also a little concerning

pkg/signaling/client.go
SimonBrandner: `RunSync()` makes it sound like it runs sync only once and then stops which is not the case - it starts syncing and keeps doing that until it encounters an error which makes me think `StartSyncing` (or `SyncUntilError`) would be better

pkg/peer/webrtc.go
SimonBrandner: Ah, yeah, that sounds right, could we get a comment here to explain why we're not sending anything here?

SimonBrandner:
> @SimonBrandner,
> 
> > This is largely looking good to me though I sometimes struggle a bit not to get lost... Perhaps a better file naming would be a bit, some of the file names aren't very descriptive from my POV.
> 
> Is there something that could be improved? I want to understand what makes you feel lost. Or is it just because the PR is too large?

The PR being large is definitely a part of it but I feel like the names don't tell me much in some cases...

> I'm definitely in favor of smaller PRs and I'm sorry that this PR is now that big: it's as big as it currently is since it's simply a complete rewrite of a previous implementation (I barely touched `main` and profiling things, but the rest was changed drastically).
> 
> Perhaps it would be easier if I published it as a new repository? (since 90% of things anyway changed) Then it would be easier to review it in an editor (or an IDE) as **a new SFU** rather than trying to comprehend how the new changes relate to the previous implementation. (which I agree must be confusing)
> 
> What I normally do for the large PRs is to split it into smaller "atomic" commits where each commit changes a single thing and explains why it was changed: that's the approach I'm in favor of and that I try to use in all projects (I've noticed that in Matrix and Element it's not always followed: many commits just alter the state of a previous one or combine unrelated things in one commit: something that I want to change!), but I think if I could present it not as one big change, but as series of well-separated commits, it would be much easier to understand (I normally rebase commits before a push and make sure they are clear, but unfortunately this time I published it earlier than I planned so that Dave can take a look at it earlier as we spoke about it).

No worries about the PR being large, no reason to create a separate repo, I realize it's a large refactor and so there aren't many alternatives

> > I am also a little concerned about completely removing the subscriber/publisher model which is very useful for simulcast and possibly other things; it's also very similar to what ion-sfu and livekit does, iirc
> 
> Good point, expected you to ask about it :)
> 
> So, the subscriber and publisher model does exist in the current implementation, but it does not have a separate module or file for it since **for our current implementation** it does not seem to be useful.
> 
> I.e. I can create a function in a `Conference` and name it `getSubscribers()` (it's easy to create one), but then it would immediately pose questions like **what exactly the subscriber structure should contain**, i.e. should it be a structure that contains the ID of a subscriber and the list of IDs of other participants to which the subscriber is subscribed to? Or should it rather be the ID of a subscriber (the participant ID) and the list of tracks? Or should it be the list of tracks and the list of participants to which we're subscribed? Etc... Currently, these can't be answered simply because we don't really need **an explicit** entity such as `Subscriber` or `Publisher`.
> 
> We may indeed need it in the future (e.g. when implementing simulcast)! And once we need it, we definitely give it a good thought and introduce it! But adding these entities right now seemed like a "premature optimization" to me (since it would pose the questions listed above) that would complicate our current (pretty trivial and simple) use case.
> 
> As for the other SFUs, I'm only familiar with some internals of Janus and LiveKit. You mentioned LiveKit: yes, they have **a concept** of a subscription and publishing a track (but that concept also we have in our implementation slightly_smiling_face), but they, like we, don't have special files or packages for `subscriber.go` or `publisher.go`, also there are no such structures as `Subscriber` or `Publisher` in their implementation. They do have a more concrete entity though like `SubscribedTrack` or `Downtrack`.
> 
> So I would not say that we don't have a publisher-subscriber model. In fact, we do have it between peers and a conference (peers publish the messages to which the conference is subscribed). And we also know which tracks everyone is subscribed it and which tracks are being published by each participant (see `participant.go`). We just do it different compared to the implementation in `main` branch - I could not inherit the Publisher/Subscriber logic from main since it unfortunately suffers from race conditions in multiple places and the entities are not separated there, so it also has bugs like a double-subscription (e.g. check `Publisher::Subscribe()` in the `main`). So they had to be reworked anyway [unfortunately].
> 
> And I'm pretty sure that with time we'll also have a more elaborated logic for this with more structures that have a publisher or subscriber-specific logic inside.

Alright, let's leave it out for now and re-add it with simulcast

> > I would also like my PR to get merged before this and for us to somehow integrate it into this one
> 
> Would you like me to integrate the logic from the RTCP PR into this PR before merging it into `main`? Or would you prefer to add it to this PR?
> 
> I can do it, but I was wondering if you would like to do it yourself to get a feel of what the rewritten version looks like and how such things are implemented within the new architecture.

Yeah, I actually wanted to have a go at adding it to this PR exactly for that reason :D Might take a bit of time, but I would like to have a look tomorrow

> > All of the TODO are also a little concerning
> 
> Yep, but there are also TODOs and FIXMEs in `main` :) There is only one place (a FIXME) in this PR that talks about a feature that is implemented in `main`, but not here. The rest existed in the previous implementation (it's just some of them were not marked as concerns and did not have a dedicated TODO, like for instance handling the data channel being busy or closing the SFU once `Sync()` ends). I decided to mark these places with TODOs (if it's not critical) and FIXMEs (if it's more critical and must be discussed), so that we don't forget about them.

Sure, can we just check we're not removing any functionality with this refactor (e.g. keep-alive)?

pkg/conference/processor.go
daniel-abramov: Ok, done. Ptal: https://github.com/matrix-org/waterfall/pull/52/commits/28296825506ec805edf6c896217efaa2034852c2 (btw, that's how I normally prefer commits to be done, i.e. "atomic" clean commits with a structured format and that do only one thing; ideally PRs should consist of such commits IMHO [like it's done in Linux kernel repository])

pkg/signaling/client.go
SimonBrandner: `RunSyncing` sounds good

daniel-abramov: Ok, renamed!

pkg/conference/participant.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Participant represents a participant in the conference.
+type Participant struct {
+	id              ParticipantID
+	logger          *logrus.Entry
+	peer            *peer.Peer[ParticipantID]
+	remoteSessionID id.SessionID
+	streamMetadata  event.CallSDPStreamMetadata
+	publishedTracks map[event.SFUTrackDescription]*webrtc.TrackLocalStaticRTP
```
SimonBrandner, line 29: Could we not use `event.SFUTrackDescription` here, please? It doesn't feel good that the Matrix event types and internal types being coupled if they aren't exactly the same thing. Also, `event.SFUTrackDescription` optionally  includes resolution, so it just feels wrong to use it as a key here since we'll never want to use resolution in the key

pkg/conference/processor.go
SimonBrandner: As discussed in our DM, we only need an `Offer` event and can get rid of `Publish` and `Unpublish` in the future

pkg/conference/participant.go
daniel-abramov: In general, I totally agree with it. I think the only reason I used it like this is that it was not clear what exactly we are going to use [as a track identifier](https://github.com/matrix-org/matrix-spec-proposals/pull/3898#discussion_r1038568858), but I agree that having a separate identifier for tracks makes more sense (it would also simplify the adjacent functions that compare things).

SimonBrandner: Yep, let's merge this ASAP - have made https://github.com/matrix-org/waterfall/issues/56. I might actually have a look into that myself if I get a change before you

pkg/common/channel.go
```diff
@@ -0,0 +1,70 @@
...
+	receiver := Receiver[M]{channel, closed}
+	return sender, receiver
+}
+
+// Sender counterpart of the channel.
+type Sender[M any] struct {
+	// The channel itself.
+	channel chan<- M
+	// Atomic variable that indicates whether the channel is closed.
+	receiverClosed *atomic.Bool
```
dbkr, line 28: My understanding is that this is considered a bit of a code smell in Go, since a sender should never be trying to write to a closed channel. I guess this means we have something where we can't know this and need to check each time?

pkg/common/message_sink.go
```diff
@@ -0,0 +1,61 @@
...
+// At the same it guarantees that the caller can't alter the `sender`, which means that
+// the sender can't impersonate another sender (and we guarantee this on a compile-time).
+type MessageSink[SenderType comparable, MessageType any] struct {
+	// The sender of the messages. This is useful for multiple-producer-single-consumer scenarios.
+	sender SenderType
+	// The message sink to which the messages are sent.
+	messageSink chan<- Message[SenderType, MessageType]
+	// Atomic variable that indicates whether the message sink is sealed.
+	// This is used to prevent sending messages to a sealed message sink.
+	// The variable is atomic because it may be accessed from multiple goroutines.
```
dbkr, line 20: For clarity's sake, could we define what 'sealed' means here?

pkg/conference/participant.go
```diff
@@ -0,0 +1,51 @@
...
+		DeviceID:        p.id.DeviceID,
+		CallID:          p.id.CallID,
+		RemoteSessionID: p.remoteSessionID,
+	}
+}
+
+func (p *Participant) sendDataChannelMessage(toSend event.SFUMessage) {
+	jsonToSend, err := json.Marshal(toSend)
+	if err != nil {
+		p.logger.Error("Failed to marshal data channel message")
```
dbkr, line 52: shouldn't this be returning here, otherwise we'll send blank messages?

SimonBrandner (approved):

pkg/common/message_sink.go
daniel-abramov: Good point, will add a comment about that! Perhaps I need to rewrite the comment for this variable altogether.

pkg/conference/participant.go
daniel-abramov: Good catch, thanks! Will be fixed right away.

pkg/common/channel.go
daniel-abramov: Good that you've noticed it: tbh I did not like the part of writing this particular structure as it felt like it's not very elegant, but unfortunately I did not find a better way to write it.

The problem that I'm trying to solve is to indeed not write to the closed channel, **but** the problem is that from Go's standpoint the channel is not closed since the channels could only be closed in Go from the sender's side, but not from the receiver's side.

A practical example where it happens in our code:
- There is a router in the code: a structure that simply has a map of all conferences running on the SFU. The router receives signaling messages (To-Device messages), checks their `conf_id,` and simply sends the message to the conference that is responsible for this message. So there is a channel between the router and a particular conference.
- Conference is listening to the channel and reacting to the messages. But at some point, all participants leave the conference which means that the conference must be ended. So the conference grouting and main loop end, sending a message to the router telling "Ok, I'm done, you can remove it".

The problem occurs because at the moment the conference is considered ended, the channel remains open (the conference can't close because the conference only holds a receiver part of the channel). And in go, there is no way to check if the channel is alive or if there is someone listening on the channel. So this means that the router might have sent certain messages to the conference expecting that the conference will read them whereas in reality, the conference has stopped listening on the channel by the moment the router gets to know that the conference is dead.

To sum up:
- There is a way in Go to check if the channel is closed by the sender on the receiver's side.
- There is no way in Go to check if there is any listener on the sender's side.
- We need a way to know (on the sender side) if someone is listening to the channel (in order to know if it makes sense to post a message).

That's pretty much the problem that I tried to solve with this logic, i.e. create a wrapper around the channel where I could inform the sender that the receiver is not listening to the changes, so that if the sender tries to send the message to such a channel, it gets the message back and knows that there are no receivers anymore.

dbkr: Thanks for the detailed explanation - I've also found Go is a bit light on actually explaining how you would write correct & non-racy code sticking to the given paradigms.

daniel-abramov: Right! I've also spent some time trying to understand how to do things safely. In Rust, for instance, we always know if there is a sender and if there is a receiver, but in Go, an elegant implementation of this does not work because of garbage collection: the receivers are not deleted immediately after leaving the scope of a function, but only get removed when the GC kicks in which means that the lifetime of the objects is not strictly defined, hence the consequence of us not knowing when the other counterpart of a channel is dead.

PR #54, opened 2022-12-03
Implement RTCP forwarding in the refactored version of the SFU

SimonBrandner:


pkg/peer/peer.go
```diff
@@ -22,6 +24,8 @@ var (
 	ErrCantSubscribeToTrack       = errors.New("can't subscribe to track")
 )
 
+const minimalPLIInterval = time.Millisecond * 500
```
SimonBrandner, line 27: Perhaps there is a better place to put this?

pkg/peer/peer.go
```diff
@@ -98,17 +102,69 @@ func (p *Peer[ID]) SubscribeTo(track *webrtc.TrackLocalStaticRTP) error {
...
 
+func (p *Peer[ID]) WriteRTCP(packets []rtcp.Packet, streamID string, trackID string, lastPLITimestamp int64) {
+	packetsToSend := []rtcp.Packet{}
+	var mediaSSRC uint32
+	for _, receiver := range p.peerConnection.GetReceivers() {
+		if receiver.Track().ID() == trackID && receiver.Track().StreamID() == streamID {
+			mediaSSRC = uint32(receiver.Track().SSRC())
+			break
+		}
+	}
```
daniel-abramov, line 130: We probably want to return an error if we have not found a track with the given `trackID` and `streamID` (otherwise also the `WriteRTCP` below would fail (unknown track).

pkg/peer/messages.go
```diff
@@ -38,3 +39,15 @@ type DataChannelMessage struct {
 }
 
 type DataChannelAvailable struct{}
+
+type ForwardRTCP struct {
```
daniel-abramov, line 43: I would probably name it as `RTCPReceived` or something. After all, these are messages, not commands that the `Peer` sends, i.e. all other commands read like "Hey, I've got a new ICE candidate { candidate }` or "Hey, I've published a new track { track }", so I think it would make sense to call this message also in a similar way to keep consistency (and then it's up to the receiver to decide what to do with them, e.g. forward, or maybe process them in a different way depending on the application logic).

pkg/peer/peer.go
SimonBrandner: Good point!

PR #69, opened 2022-12-09
conference: treat `Publish` and `Unpublish` equally

### LINKED ISSUE #68: Ending screen-sharing doesn't generate a new SDP offer being send to other clients
1. Have a call with two clients (A and B)
2. A start screen-sharing
3. A stops screen-sharing
4. B sees stuck screen-sharing feed which isn't changing
5. A leaves the call
6. B sees both feeds disappear

daniel-abramov:
Fixes https://github.com/matrix-org/waterfall/issues/68.


SimonBrandner (approved):
Yep, that makes sense!

(/me hopes to get the event shape changes done soonish)

PR #70, opened 2022-12-10
Change event shape to match MSC3898

### LINKED ISSUE #62: Update events shape to match the MSC

SimonBrandner:
Fixes https://github.com/matrix-org/waterfall/issues/62
Requires https://github.com/matrix-org/matrix-js-sdk/pull/2964


pkg/conference/config.go
```diff
@@ -2,7 +2,11 @@ package conference
...
 type Config struct {
-	// Keep-alive timeout for WebRTC connections. If no keep-alive has been received
-	// from the client for this duration, the connection is considered dead (in seconds).
+	// Keep-alive timeout for WebRTC connections. If the client doesn't respond
+	// to an `m.call.ping` with an `m.call.pong` for this amount of time, the
+	// connection is considered dead. (in seconds, no greater then 30)
 	KeepAliveTimeout int `yaml:"timeout"`
+	// The time after which we should send another m.call.ping event to the
+	// client. (in seconds, greater then 30)
+	PingInterval int `yaml:"pingInterval"`
```
dbkr, line 11: Making this configurable is probably overkill and will give people another way to shoot themselves in the foot, but we can always comment it as, "you really don't want to change this" in the config.

pkg/conference/data_channel_message_processor.go
```diff
@@ -28,47 +36,61 @@ func (c *Conference) processSelectDCMessage(participant *Participant, msg event.
...
+func (c *Conference) processNegotiateDCMessage(participant *Participant, msg event.FocusCallNegotiateEventContent) {
+	participant.streamMetadata = msg.SDPStreamMetadata
 
-	answer, err := participant.peer.ProcessSDPOffer(msg.SDP)
-	if err != nil {
-		participant.logger.Errorf("Failed to set SDP offer: %v", err)
-		return
-	}
+	if msg.Description.Type == event.CallDataTypeOffer {
+		participant.logger.Info("Received SDP offer over DC")
```
dbkr, line 55: Can we log the offer itself at trace level?

pkg/config/config.go
```diff
@@ -77,7 +77,9 @@ func LoadConfigFromString(configString string) (*Config, error) {
 	if config.Matrix.UserID == "" ||
 		config.Matrix.HomeserverURL == "" ||
 		config.Matrix.AccessToken == "" ||
-		config.Conference.KeepAliveTimeout == 0 {
+		config.Conference.KeepAliveTimeout == 0 ||
+		config.Conference.KeepAliveTimeout > 30 ||
+		config.Conference.PingInterval < 30 {
```
dbkr, line 82: At some point I'd suggest splitting these out into multiple checks so we can print an error message that tells you what's wrong, otherwise this could get unwieldy, but doesn't have to be now.

dbkr (approved):
Otherwiswe lgtm

pkg/config/config.go
SimonBrandner: Have added a TODO and made https://github.com/matrix-org/waterfall/issues/71

pkg/conference/config.go
SimonBrandner: Well, we validate the config, so I think we don't have to worry about it much?

SimonBrandner: True... Will add some warning too when adding the config errors

SimonBrandner: (https://github.com/matrix-org/waterfall/issues/71)

daniel-abramov, line 11: NB: If we're already extending the configuration for the keepalive/heartbeat logic, would it make sense to group them together in their own struct: `KeepAliveConfig` ([though keep-alive is possibly not the best term here](https://github.com/matrix-org/matrix-spec-proposals/pull/3898#discussion_r1042049466)) or `HeartBeatConfig`? So that it's obvious that both fields actually go together (it would also mean, we could supply this structure with grouped/related fields directly to the one who needs it instead of passing fields one-by-one [passing one-by-one can mix the order due to typos etc and also might be less obvious])

pkg/conference/data_channel_message_processor.go
```diff
@@ -28,47 +36,61 @@ func (c *Conference) processSelectDCMessage(participant *Participant, msg event.
 		}
 	}
 
-	// Subscribe to the found tracks.
+	// Subscribe to the found tracks
```
daniel-abramov, line 39: OCD: It seems like all other comments have a full-stop at the end, but it was removed for this comment for some reason. Could we get it back? 🙂

pkg/conference/data_channel_message_processor.go
```diff
@@ -28,47 +36,61 @@ func (c *Conference) processSelectDCMessage(participant *Participant, msg event.
...
-}
-
-func (c *Conference) processAnswerDCMessage(participant *Participant, msg event.SFUMessage) {
-	participant.logger.Info("Received SDP answer over DC")
 
-	if err := participant.peer.ProcessSDPAnswer(msg.SDP); err != nil {
-		participant.logger.Errorf("Failed to set SDP answer: %v", err)
-		return
-	}
+	// TODO: Handle unsubscribe
```
daniel-abramov, line 48: Ah, so we don't want to implement it right now? Could we probably move this TODO a bit higher in a function (maybe before handling the `subscribe`?), so that we don't forget about it (it would also influence the `subscribe` handling when e.g. `unsubscribe` would contain tracks that are also mentioned in `subscribe` which is a bug on a client side, but must be handled by the server properly)

pkg/conference/data_channel_message_processor.go
```diff
@@ -28,47 +36,61 @@ func (c *Conference) processSelectDCMessage(participant *Participant, msg event.
...
-	participant.logger.Info("Received SDP offer over DC")
+func (c *Conference) processNegotiateDCMessage(participant *Participant, msg event.FocusCallNegotiateEventContent) {
+	participant.streamMetadata = msg.SDPStreamMetadata
 
-	answer, err := participant.peer.ProcessSDPOffer(msg.SDP)
-	if err != nil {
-		participant.logger.Errorf("Failed to set SDP offer: %v", err)
-		return
-	}
+	if msg.Description.Type == event.CallDataTypeOffer {
```
daniel-abramov, line 54: We have `if`, `else if`, and `else` for checking the type of the event here. It sounds like a good candidate for a `switch` statement to me. It feels like it would be a bit more elegant to switch for the type (good use case for a `switch`).

pkg/conference/matrix_message_processor.go
```diff
@@ -53,7 +53,21 @@ func (c *Conference) onNewParticipant(participantID ParticipantID, inviteEvent *
...
+				Content: event.Content{},
+			})
+		}
+		peer, answer, err := peer.NewPeer(
+			inviteEvent.Offer.SDP,
+			messageSink,
+			logger,
+			pingInterval,
+			keepAliveDeadline,
+			sendPing,
```
daniel-abramov, line 69: It seems like we're passing 3 parameters that are only related to the keep-alive/heartbeat logic. I think it would probably look a bit easier to read and maintain if we maybe group them together in a structure or something.

I also see that now we construct one of the closures inside the `Peer` as it was before, but another one inside `Conference`. Maybe in this case it would make sense to move the keep-alive handling to the conference level.

Or, pass the `HeartbeatConfiguration` or something like this to the `Peer`, so that the `Peer` just starts a go-routine passing a single parameter to the otherwise opaque structure? This would also mean we would not need to store keep-alive related vars inside `Peer`.

We could start the go-routine once `Peer` is constructed (like we did before) so that if the data channel is not connected within the deadline, we would treat it as hangup (also like we did before), which probably makes sense.

pkg/peer/webrtc.go
```diff
@@ -133,6 +133,14 @@ func (p *Peer[ID]) onDataChannelReady(dc *webrtc.DataChannel) {
...
 		p.logger.Info("Data channel opened")
 		p.sink.Send(DataChannelAvailable{})
+
+		startKeepAlive(
+			p.pingInterval,
+			p.keepAliveDeadline,
+			p.pong,
+			p.sendPing,
+			func() { p.sink.Send(LeftTheCall{event.CallHangupKeepAliveTimeout}) },
+		)
```
daniel-abramov, line 143: Note, we probably want to try to start this go-routine earlier (once the peer is constructed), so that the peer is thrown away if no data channels are available or if the opening of the data channel takes too long.

pkg/conference/config.go
SimonBrandner: Have added it to the list of config stuff to fix in a follow up PR: https://github.com/matrix-org/waterfall/issues/71

pkg/conference/data_channel_message_processor.go
SimonBrandner: > Ah, so we don't want to implement it right now?

It didn't feel like this should be a part of this PR. Will do a follow-up

> Could we probably move this TODO a bit higher in a function (maybe before handling the `subscribe`?), so that we don't forget about it (it would also influence the `subscribe` handling when e.g. `unsubscribe` would contain tracks that are also mentioned in `subscribe` which is a bug on a client side, but must be handled by the server properly)

Will do

pkg/conference/peer_message_processor.go
```diff
@@ -68,40 +67,55 @@ func (c *Conference) processICEGatheringCompleteMessage(participant *Participant
...
-		c.processAnswerDCMessage(participant, sfuMessage)
-	case event.SFUOperationPublish, event.SFUOperationUnpublish:
-		c.processPublishDCMessage(participant, sfuMessage)
-	case event.SFUOperationAlive:
-		c.processAliveDCMessage(participant)
-	case event.SFUOperationMetadata:
-		c.processMetadataDCMessage(participant, sfuMessage)
+	switch focusEvent.Type.Type {
+	case event.FocusCallTrackSubscription.Type:
+		focusEvent.Content.ParseRaw(event.FocusCallTrackSubscription)
```
daniel-abramov, line 94: Just a question: is there a way we could avoid calling `ParseRaw()` for every event type?

I.e. so that it's enough to call `AsXXX()` after we checked the `EventType`? (similar to how we do it with other signaling messages from e.g. Matrix SDK, where we just check the type and then call one of the `AsXXX()` functions?)

daniel-abramov (approved):

PR #74, opened 2022-12-12
Config clean-up

### LINKED ISSUE #71: Settings cleanup
- Add proper errors when the config is invalid
- Add warnings when using non-intelligent values
- Group heartbeat config

SimonBrandner:
Fixes https://github.com/matrix-org/waterfall/issues/71

### DIFF
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -3,4 +3,6 @@ matrix:
   userid: "@sfu:shadowfax"
   accesstoken: "..."
 conference:
-  timeout: 30
+  pingPong:
+    timeout: 30
+    interval: 30
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -3,6 +3,6 @@ matrix:
   userid: "@sfu:shadowfax"
   accesstoken: "..."
 conference:
-  pingPong:
+  heartbeat:
     timeout: 30
     interval: 30
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -96,31 +96,31 @@ func ValidateConfig(config Config) error {
 		errored = true
 		fmt.Println("You must set matrix.accessToken")
 	}
-	if config.Conference.PingPongConfig.Timeout == 0 {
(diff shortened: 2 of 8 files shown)


config.yaml.sample
```diff
@@ -3,4 +3,6 @@ matrix:
   userid: "@sfu:shadowfax"
   accesstoken: "..."
 conference:
-  timeout: 30
+  pingPong:
+    timeout: 30
+    interval: 30
```
daniel-abramov, line 8: Philosophical question: is there a more descriptive name that we could use for this? I know that keep-alive and heartbeat are used very often for the same meaning (the keep-alive is more used for marking connections alive, TCP's mechanism of keeping the connection alive is called like this; while heartbeat is oftentimes used for similar responsiveness checks).

[RFC 6455](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.3) e.g. uses a heartbeat term for that.

pkg/config/config.go
```diff
@@ -74,15 +74,58 @@ func LoadConfigFromString(configString string) (*Config, error) {
...
+		errored = true
+		fmt.Println("It is not recommended for pingPong.interval to be more than 300s")
+	}
+
+	if errored {
+		return errors.New("invalid config values")
+	} else {
+		return nil
+	}
+}
```
daniel-abramov, line 131: Generally `fmt.Println()` is not the best way to report errors when doing it outside of a `main` package (also it uses the stdout and may collide with whatever logging framework is used). For packages and libraries, it's more common to just return an error (the caller will unpack the error and print it if it's of significance).

I think the function will look more elegant and concise if each `if` returns a certain error. The caller (in `main()`) would anyway get this error and print it via the logging framework.

daniel-abramov (changes requested):

pkg/config/config.go
daniel-abramov: One option would be to create a couple of errors in the top of the file and them parameterize them with the field that failed or something like this, but ok, let me approve and improve it later!

daniel-abramov (approved):

PR #80, opened 2022-12-15
Fix issues with the heartbeat logic on the server

daniel-abramov:
Per-commit review is required (one fix - one comment with an explanation).

Summary:
- Fixed memory leak caused by `Ticker()` in the heartbeat.
- Fixed heartbeat not assuming that pings can fail (data channel may be closed for instance).
- Fixed wrong comparison in config validation (that mandated interval being 30+ seconds and not 30- seconds as per MSC).
- Remove confusing (duplicate) logs (and make less relevant ones debug).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -101,8 +101,8 @@ func validateConfig(config Config) error {
 	if config.Conference.HeartbeatConfig.Timeout > 30 {
 		return fmt.Errorf("heartbeat.timeout must be 30s or lower")
 	}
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -101,8 +101,8 @@ func validateConfig(config Config) error {
 	if config.Conference.HeartbeatConfig.Timeout > 30 {
```
SimonBrandner, line 101: Argh, I got them the other way around I think...

SimonBrandner (approved):

PR #81, opened 2022-12-15
Add more debugging for ending calls

SimonBrandner:


pkg/peer/webrtc.go
```diff
@@ -110,6 +110,7 @@ func (p *Peer[ID]) onConnectionStateChanged(state webrtc.PeerConnectionState) {
 
 	switch state {
 	case webrtc.PeerConnectionStateFailed, webrtc.PeerConnectionStateDisconnected, webrtc.PeerConnectionStateClosed:
+		p.logger.Debug("Ending call since we're in state: %v", state)
```
daniel-abramov, line 113: We probably don't need this one since we log the state as `Info` just 2 lines above this 🙂

pkg/conference/matrix_message_processor.go
```diff
@@ -68,6 +68,7 @@ func (c *Conference) onNewParticipant(participantID ParticipantID, inviteEvent *
 				}) == nil
 			},
 			OnTimeout: func() {
+				c.logger.Debug("We haven't received an m.call.ping from the client for more than the timeout")
```
daniel-abramov, line 71: Note, we probably also don't necessarily need to log it here, because when the `LeftTheCall` message with the keep alive timeout reason is sent in a line below, it gets logged already. Currently, the reason is `Hangup` though, so it's the `LeftTheCall` that we get from the connection state change below, not from the timeout.

SimonBrandner:
Turns out we already have sufficient logging

PR #83, opened 2022-12-16
Fix config validation to match MSC3898

SimonBrandner:
I got this the other way around the first time: https://github.com/matrix-org/matrix-spec-proposals/blob/SimonBrandner/msc/sfu/proposals/3898-sfu.md#mcallping-mcallpong

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -98,21 +98,12 @@ func validateConfig(config Config) error {
 		return fmt.Errorf("you must set heartbeat.interval")
 	}
 
-	// MSC3898 dictates what are the limits for these values.
-	if config.Conference.HeartbeatConfig.Timeout < 30 {
-		return fmt.Errorf("heartbeat.timeout must be 30s or higher")
+	// Make sure the heartbeat values are within sane bounds
+	if config.Conference.HeartbeatConfig.Timeout < 30 && config.Conference.HeartbeatConfig.Timeout > 60*2 {
+		return fmt.Errorf("heartbeat.timeout must be between 30s and 2m")
 	}
-	if config.Conference.HeartbeatConfig.Interval > 30 {
-		return fmt.Errorf("heartbeat.interval must be 30s or lower")
-	}
-
-	// While MSC3898 isn't explicit about this, we don't want users to shoot
-	// themselves in the foot.
-	if config.Conference.HeartbeatConfig.Timeout > 60*2 {
-		return fmt.Errorf("it is not recommended for heartbeat.timeout to be higher than 2m")
-	}
-	if config.Conference.HeartbeatConfig.Interval < 5 {
-		return fmt.Errorf("it is not recommended for heartbeat.interval to be less than 5s")
+	if config.Conference.HeartbeatConfig.Interval < 5 && config.Conference.HeartbeatConfig.Interval > 30 {
+		return fmt.Errorf("heartbeat.interval must be between 5s and 30s")
 	}
 
 	return nil
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -98,18 +98,21 @@ func validateConfig(config Config) error {
 		return fmt.Errorf("you must set heartbeat.interval")
 	}
 
-	if config.Conference.HeartbeatConfig.Timeout > 30 {
-		return fmt.Errorf("heartbeat.timeout must be 30s or lower")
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -98,18 +98,21 @@ func validateConfig(config Config) error {
 		return fmt.Errorf("you must set heartbeat.interval")
 	}
 
-	if config.Conference.HeartbeatConfig.Timeout > 30 {
-		return fmt.Errorf("heartbeat.timeout must be 30s or lower")
+	// MSC3898 dictates what are the limits for these values.
+	if config.Conference.HeartbeatConfig.Timeout < 30 {
+		return fmt.Errorf("heartbeat.timeout must be 30s or higher")
```
daniel-abramov, line 103: Do we really want it to be lower than 30 seconds? - If we do so, it means that the overall timeout when sending pings would be `interval + timeout` because after sending each ping we don't send a new ping before receiving a pong.

Or is it expected that we don't wait for confirmations for each pong and just shoot pings using the interval disregarding timeout? - If so, we would need to change our implementation (currently it's ping that expects an acknowledgment, i.e. pong). Another alternative is to make each side send **a unidirectional heartbeat** meaning that `ping`s are heartbeats that the server sends to the client so that the client can detect when the server is dead. While `pong`s are "client issues pings" essentially, the server expects to come each `timeout` (then we would need to rename values also).

pkg/config/config.go
```diff
@@ -98,18 +98,21 @@ func validateConfig(config Config) error {
...
 	if config.Conference.HeartbeatConfig.Interval > 30 {
 		return fmt.Errorf("heartbeat.interval must be 30s or lower")
 	}
 
-	if config.Conference.HeartbeatConfig.Timeout < 5 {
-		return fmt.Errorf("heartbeat.timeout must be 5s or higher")
+	// While MSC3898 isn't explicit about this, we don't want users to shoot
+	// themselves in the foot.
+	if config.Conference.HeartbeatConfig.Timeout > 60*2 {
+		return fmt.Errorf("it is not recommended for heartbeat.timeout to be higher than 2m")
```
daniel-abramov, line 112: We are checking the `Timeout` value above in the code, maybe it would be better to just combine the `if` condition and return an error instead with a text that "timeout is expected to be in range A..B"?

pkg/config/config.go
```diff
@@ -98,18 +98,21 @@ func validateConfig(config Config) error {
...
 	}
-	if config.Conference.HeartbeatConfig.Interval > 60*5 {
-		return fmt.Errorf("heartbeat.interval must be 5m or lower")
+	if config.Conference.HeartbeatConfig.Interval < 5 {
+		return fmt.Errorf("it is not recommended for heartbeat.interval to be less than 5s")
```
daniel-abramov, line 115: Hm, but according to the MSC:

> A ping message must be sent by the focus to the client **at an interval no greater than** 30 seconds.

SimonBrandner: While the MSC allows for such a small interval, it certainly isn't a good idea to use such a small interval since it might result in the SFU being flooded with `m.call.pong` events

pkg/config/config.go
SimonBrandner: Well, from my POV they are different errors: one is about what the MSC allows and the other is about what we recommend

pkg/config/config.go
SimonBrandner: From the MSC it's not 100% clear to me... @dbkr, what are your thoughts on this?

dbkr: Sorry for not making this super clear - it's the client side timeout that needs to be greater than 30s (because that's how often the server will be sending ping packets). The server side timeout just needs to be long enough that the server is sure the client isn't there any more - I'd suggest maybe 10 seconds?

I'd expect us to fire pings regularly and not wait for a pong response before starting the timer - that way the maximum 30s interval is maintained.

Would some suggested values in the MSC help make this clearer?

pkg/config/config.go
dbkr: I think it depends whether one is an ignorable warning & the other is an error - if so then they'd want to be separate. If they're both hard errors I'd just make them the same. I still think just hard coding the values might prevent overthinking this. :)

daniel-abramov (approved):

PR #85, opened 2022-12-20
Poor man's simulcast implementation

### LINKED ISSUE #31: Implement basic simulcast
@daniel-abramov:

- Receive multiple spatial layers on video, forward them as a single layer.
- Let the user select the layer with `subscribe` command.

No congestion control or rate allocation for now!

daniel-abramov:
Per-commit review is suggested (tried to keep the commit history clean and "atomic").

As discussed, I'm submitting this PR to simplify the review of the simulcast that would otherwise have been too overwhelming (over 1000 LOC PR).

This is stage 1 of simulcast implementation (fixes https://github.com/matrix-org/waterfall/issues/31).


pkg/conference/data_channel_message_processor.go
```diff
@@ -11,12 +13,59 @@ func (c *Conference) processTrackSubscriptionDCMessage(
...
+		toSubscribeTrackIDs = append(toSubscribeTrackIDs, track.TrackID)
+		toSubscribeRequirements[track.TrackID] = TrackMetadata{track.Width, track.Height}
+	}
+
+	// Calculate the list of tracks we need to subscribe and unsubscribe from based on the requirements.
+	subscribeTo := []peer.ExtendedTrackInfo{}
+
+	// Iterate over all published tracks that correspond to the track IDs we want to subscribe to.
+	for id, track := range c.getTracksInfo(toSubscribeTrackIDs) {
+		// Get tre tracks we're subscribed to.
```
dbkr, line 35:
```suggestion
		// Get the tracks we're subscribed to.
```

dbkr (approved):

pkg/peer/webrtc_callbacks.go
```diff
@@ -11,44 +10,33 @@ import (
...
-	localTrack, err := webrtc.NewTrackLocalStaticRTP(
-		remoteTrack.Codec().RTPCodecCapability,
-		remoteTrack.ID(),
-		remoteTrack.StreamID(),
-	)
+	// Get the simulcast layer from RID.
+	layer, err := RIDToSimulcastLayer(remoteTrack.RID())
 	if err != nil {
-		p.logger.WithError(err).Error("failed to create local track")
+		p.logger.WithError(err).Error("failed to parse RID")
```
SimonBrandner: We should handle non-simulcast tracks too, right?

pkg/conference/peer_message_processor.go
```diff
@@ -19,27 +19,46 @@ func (c *Conference) processLeftTheCallMessage(participant *Participant, msg pee
...
 
-	participant.publishedTracks[msg.Track.ID()] = PublishedTrack{track: msg.Track}
+	// TODO: Oops, that's not very efficient with the simulcast since it means that when 3
+	// layers are published, we will send 3 times the same metadata.
 	c.resendMetadataToAllExcept(participant.id)
 }
 
+func (c *Conference) processRTPPacketReceivedMessage(participant *Participant, msg peer.RTPPacketReceived) {
+	// For now we just forward the lowest layer always and assume that others don't exist.
+	if msg.Layer != peer.SimulcastLayerLow {
```
SimonBrandner: Shouldn't it be this instead? Won't the current version forward all but the low layer?

```suggestion
	if msg.Layer == peer.SimulcastLayerLow {
```

pkg/peer/types.go
```diff
@@ -47,15 +60,21 @@ func (s SimulcastLayer) String() string {
...
-type SimulcastTrackInfo struct {
+// Information with some extended (optional fields).
+// Ideally we would want to have different types for different tracks, but Go does not have ADTs,
+// so it's the only way to do it without writing too much of a boilerplate.
+type ExtendedTrackInfo struct {
+	// Information about a track.
 	TrackInfo
-	Layer SimulcastLayer
+	// Optional simulcast layer (if any). Would be nil for non-simulcast or audio tracks.
+	Layer *SimulcastLayer
```
SimonBrandner: I don't like the usage of a pointer here... Could we add `SimulcastLayer.Invalid` or something along those lines?

pkg/conference/peer_message_processor.go
```diff
@@ -19,32 +19,47 @@ func (c *Conference) processLeftTheCallMessage(participant *Participant, msg pee
...
+		return
 	}
 
-	// TODO: Oops, that's not very efficient with the simulcast since it means that when 3
-	// layers are published, we will send 3 times the same metadata.
-	c.resendMetadataToAllExcept(participant.id)
+	// If it's just a new layer, let's add it to the list of layers of the existing published track.
+	if msg.Layer != nil && !funk.Contains(track.layers, *msg.Layer) {
+		track.layers = append(track.layers, *msg.Layer)
+		participant.publishedTracks[msg.TrackID] = track
```
SimonBrandner, line 44: Won't all the layers have the same `trackId`?

pkg/conference/participant.go
```diff
@@ -64,3 +64,77 @@ func (p *Participant) sendDataChannelMessage(toSend event.Event) error {
...
+			metadata:        metadata,
+		}
+	}
+
+	return publishedTracksMetadata
+}
+
+func (p *PublishedTrackInfo) getDesiredLayer(requestedWidth, requestedHeight int) *peer.SimulcastLayer {
+	// Audio track. For them we don't have any simulcast.
+	if p.metadata.isVideoTrack() {
```
SimonBrandner:
```suggestion
	if !p.metadata.isVideoTrack() {
```

pkg/conference/participant.go
```diff
@@ -64,3 +64,77 @@ func (p *Participant) sendDataChannelMessage(toSend event.Event) error {
...
+	}
+
+	// Video track. Calculate it's full resolution based on a metadata.
+	fullResolution := p.metadata.fullResolution()
+
+	// If no explicit resolution specified, subscribe to the lowest layer.
+	desiredLayer := peer.SimulcastLayerLow
+
+	// Determine which simulcast desiredLayer to subscribe to based on the requested resolution.
+	if requestedWidth != 0 || requestedHeight != 0 {
```
SimonBrandner, line 132: Otherwise one could be `0` and therefore `desiredResolution` would too be `0`

```suggestion
	if requestedWidth != 0 && requestedHeight != 0 {
```

pkg/peer/peer.go
```diff
@@ -246,3 +217,32 @@ func (p *Peer[ID]) ProcessSDPOffer(sdpOffer string) (*webrtc.SessionDescription,
...
+			case *rtcp.PictureLossIndication:
+				toForward = append(toForward, PictureLossIndicator)
+			case *rtcp.FullIntraRequest:
+				toForward = append(toForward, FullIntraRequest)
+			}
+		}
+
+		p.sink.Send(RTCPReceived{Packets: toForward, TrackID: rtpSender.Track().ID()})
+	}
+}
```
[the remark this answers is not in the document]
daniel-abramov: I think I might have moved it just to keep the order of functions as "first public functions/interface, then private functions", i.e. `WriteRTCP()` is a public function (Go is case sensitive: functions, types, fields that start with capital are exported, whereas those written with a lowercase are internal to the package [folder]).

I should have done it in a separate commit explaining this reasoning to avoid confusion.

pkg/peer/webrtc_api.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+
+	"github.com/pion/interceptor"
+	"github.com/pion/webrtc/v3"
+)
+
+// Creates Pion's WebRTC API that has all required extensions configured (such as simulcast).
+func createWebRTCAPI() (*webrtc.API, error) {
+	// Enable Extension Headers needed for Simulcast.
```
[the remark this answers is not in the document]
daniel-abramov: Oh, you're right, going to remove it!

pkg/peer/webrtc_api.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, fmt.Errorf("failed to set default interceptors: %w", err)
+	}
+
+	return webrtc.NewAPI(webrtc.WithMediaEngine(mediaEngine), webrtc.WithInterceptorRegistry(interceptor)), nil
+}
+
+// Creates a peer connection with a specifically configured API (with simulcast etc).
+func createPeerConnection() (*webrtc.PeerConnection, error) {
+	// TODO: Could we actually reuse the same API configuration for all peer connections?
+	api, err := createWebRTCAPI()
```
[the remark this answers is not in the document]
daniel-abramov: Good question: that's why I added the TODO to clarify that :)

The previous implementation from the [draft PR](https://github.com/matrix-org/waterfall/pull/33/files#diff-e83b4d763d27611915a497672c603413fde08249808a76dc627ca661db39c46fR346) created it for each peer connection, so I thought that there must be a reason for that. But if you say that re-using is fine (you probably played more with this API than I did 🙂), then I would change it to re-use the same API.

pkg/peer/webrtc_callbacks.go
daniel-abramov: Yep, I've added this behavior in the upcoming commits (perhaps something went wrong with the rebase), but you probably have already noticed it in the newer commits :)

pkg/conference/peer_message_processor.go
daniel-abramov: Yes, the version of that commit indeed only forwarded the low layer and did not allow switching (so in the first iteration I just wrote some sort of a "drop-in" replacement), in the upcoming commit (poor man's implementation) I handled all layers.

pkg/conference/participant.go
daniel-abramov: Thanks! Updated.

pkg/conference/peer_message_processor.go
daniel-abramov: They will, that's why in this case I only modify the `layers` without touching the rest of the fields.

pkg/conference/participant.go
daniel-abramov: Yep, was indeed a bug in the initial commit, [fixed it by accident in this commit](https://github.com/matrix-org/waterfall/pull/85/commits/59579be99a6879efbb6a342d2ed864d5147f596b).

pkg/peer/types.go
daniel-abramov: Yes, that's what I did here: https://github.com/matrix-org/waterfall/pull/85/commits/59579be99a6879efbb6a342d2ed864d5147f596b

Although it's a kludge for missing ADTs, it seems like in Go they oftentimes use pointers to signify such things.

pkg/peer/types.go
```diff
@@ -0,0 +1,97 @@
...
+	Codec    webrtc.RTPCodecCapability
+}
+
+// Information with some extended (optional fields).
+// Ideally we would want to have different types for different tracks, but Go does not have ADTs,
+// so it's the only way to do it without writing too much of a boilerplate.
+type ExtendedTrackInfo struct {
+	// Information about a track.
+	TrackInfo
+	// Optional simulcast layer (if any). Would be nil for non-simulcast or audio tracks.
```
SimonBrandner, line 75: Didn't we want to use `SimulcastLayerNone` here?

pkg/peer/types.go
```diff
@@ -0,0 +1,97 @@
...
+type TrackInfo struct {
+	TrackID  string
+	StreamID string
+	Codec    webrtc.RTPCodecCapability
+}
+
+// Information with some extended (optional fields).
+// Ideally we would want to have different types for different tracks, but Go does not have ADTs,
+// so it's the only way to do it without writing too much of a boilerplate.
+type ExtendedTrackInfo struct {
```
SimonBrandner, line 72: Is there a reason for this to be separate from `TrackInfo`?

pkg/conference/participant/track.go
```diff
@@ -0,0 +1,100 @@
...
+	Owner ID
+	// Info about the track.
+	Info peer.TrackInfo
+	// Available simulcast Layers.
+	Layers []peer.SimulcastLayer
+	// Track metadata.
+	Metadata TrackMetadata
+	// The timestamp at which we are allowed to send the FIR or PLI request. We don't want to send them
+	// too often, so we introduce some trivial rate limiting to not "enforce" too many key frames.
+	canSendKeyframeAt time.Time
```
SimonBrandner, line 24:
```suggestion
	canRequestKeyframeAt time.Time
```

pkg/conference/participant/track.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// If we got here, then the low layer was requested, but it's not available.
+	// Let's try to return medium then if it's available.
+	if funk.Contains(p.Layers, peer.SimulcastLayerMedium) {
+		return peer.SimulcastLayerMedium
+	}
+
+	// No other choice rather than sending low.
+	return peer.SimulcastLayerLow
```
SimonBrandner, line 85: I wonder if this logic could be somehow simplified in a for loop, I know I had something along those lines in my original PR, not sure how well it worked/was readable :thinking:

pkg/conference/participant/tracker.go
```diff
@@ -0,0 +1,231 @@
...
+	"fmt"
+	"time"
+
+	"github.com/matrix-org/waterfall/pkg/peer"
+	"github.com/pion/rtp"
+	"github.com/thoas/go-funk"
+)
+
+type (
+	Subscribers      map[ID]*peer.Subscription
```
SimonBrandner, line 13:
```suggestion
	Subscriptions      map[ID]*peer.Subscription
```

pkg/conference/participant/tracker.go
```diff
@@ -0,0 +1,231 @@
...
+			t.RemoveTrack(trackID)
+		}
+	}
+
+	return streamIdentifiers
+}
+
+// Adds a new track to the list of published tracks, i.e. by calling it we inform the tracker that there is new track
+// that has been published and that we must take into account from now on.
+func (t *Tracker) AddTrack(
```
SimonBrandner, line 82:
```suggestion
func (t *Tracker) PublishTrack(
```

SimonBrandner (commented):
I have to admit that some of the abstractions are becoming a bit unclear to me again (e.g. `Subscription` vs `SubscribedTrack`). I wonder if the naming could be improved here

pkg/peer/types.go
daniel-abramov: We do. I think I need to fix the outdated comment 🙂

pkg/peer/types.go
daniel-abramov: Hm, good question, I've noticed that `ExtendedTrackInfo` is used in most places. Let me get rid of this abstraction then!

pkg/conference/participant/track.go
daniel-abramov: NB: I just took the name of this field from the previous implementation 🙂 Though I can definitely rename it, requesting probably reflects its meaning better.

pkg/conference/participant/track.go
daniel-abramov: Good point: I also don't like this `if` boilerplate here. Previously it was a single `if`, it changed when I realized that the layers can be published in all the combinations: high + low, high + medium, medium + log, high only, etc, handling all of them is not super elegant.

My original logic was that we respect what the user sent to us when requesting the track and use the layer that is desired by the user. However, if a given layer is not available, we must not automatically switch to a different one (higher or lower), but report an error to the user (so that the user can decide if he wants to have a higher layer e.g. if no other layers are available or rather prefer to not have any video at all [for one or another reason]).

pkg/conference/participant/tracker.go
daniel-abramov: But we don't publish the track here: the way has been posted at this point already, we add it to the tracker, so we start tracking its state and keep it in sync with the participants' state. Another reason why I chose `AddTrack()` here is to be consistent and symmetric with other functions: we have `add participant`, `remove participant`, `for each participant` and we also have `add track`, `remove track` and `for each track`.

Otherwise having only `add track` called `publish track` might be confusing (i.e. why `publish track`, but `remove track` etc).

daniel-abramov:
> I have to admit that some of the abstractions are becoming a bit unclear to me again (e.g. Subscription vs SubscribedTrack). I wonder if the naming could be improved here

We don't have a `SubscribedTrack` structure if I'm not mistaken 🤔, but we do have a `Subscription` which refers to a single subscription (subscription to a single track).

Or did you mean that you would prefer `Subscription` called `SubscribedTrack`? - I thought that it might be confusing because it would mean that `SubscribedTrack` encapsulates a track to which someone is subscribed (to me it feels like it gives a notion that there is one `SubscribedTrack`, i.e. if `asdf` is a track, to which someone is subscribed, does it have to be a type of `SubscribedTrack` and stored in a map, so that we can differentiate tracks to which someone is subscribed vs tracks that no-one is subscribed to).

Whereas with `Subscription` it's just a **subscription** (has a connotation that it is a **new** entity) to an _existing_ published track, which means that there could be many of them, i.e. we may have 3,4 or 10 subscriptions to a single existing track.

pkg/conference/participant/tracker.go
SimonBrandner: Right, what about `AddPublishedTrack()`? I wanted there to be a clear distinction between subscribing and adding a published track

SimonBrandner:
> > I have to admit that some of the abstractions are becoming a bit unclear to me again (e.g. Subscription vs SubscribedTrack). I wonder if the naming could be improved here
> 
> We don't have a `SubscribedTrack` structure if I'm not mistaken thinking, but we do have a `Subscription` which refers to a single subscription (subscription to a single track).
> 
> Or did you mean that you would prefer `Subscription` called `SubscribedTrack`? - I thought that it might be confusing because it would mean that `SubscribedTrack` encapsulates a track to which someone is subscribed (to me it feels like it gives a notion that there is one `SubscribedTrack`, i.e. if `asdf` is a track, to which someone is subscribed, does it have to be a type of `SubscribedTrack` and stored in a map, so that we can differentiate tracks to which someone is subscribed vs tracks that no-one is subscribed to).
> 
> Whereas with `Subscription` it's just a **subscription** (has a connotation that it is a **new** entity) to an _existing_ published track, which means that there could be many of them, i.e. we may have 3,4 or 10 subscriptions to a single existing track.

Ah, yes, the whole `Subscription`, `Subscribers` and `TrackSubscribers` has me confused...

pkg/conference/participant/track.go
daniel-abramov: Done.

pkg/conference/participant/tracker.go
daniel-abramov: Ah, but there is no function to mix it with subscribing, i.e. for subscribing there is a `Subscribe` function in a tracker, so adding tracks (`AddTrack()`) is only for published tracks (which we also explain in a comment).

I renamed it to `PublishedTrack` for each function, i.e. `AddTrack` is now `AddPublishedTrack`, and `UpdateTrackMetadata` is now `UpdatePublishedTrackMetadata`. Let me know if it's better (to me it seems a bit too lengthy now, but I don't have a strong opinion on that 🙂).

pkg/peer/types.go
daniel-abramov: Done.

One side effect that I noticed while doing it (and which also reminded me why I introduced it in the first place) is that `PublishedTrack` that contains `TrackInfo` **and** `Layers` (list of published simulcast layers) is slightly confusing in a way that `publishedTrack.Info.Layer` is a useless variable now. But maybe we can ignore it for now as we don't really use it apart single place (which you'll see in a commit that introduced the change). Apart from that it indeed looks simpler.

pkg/common/track_info.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+func (c ConnectionWrapper) Subscribe(track *webrtc.TrackLocalStaticRTP) (*webrtc.RTPSender, error) {
+	return c.connection.AddTrack(track)
+}
+
+func (c ConnectionWrapper) Unsubscribe(sender *webrtc.RTPSender) error {
+	return c.connection.RemoveTrack(sender)
+}
```
SimonBrandner, line 95: This looks like it should be in a different file? Also, should the `c`s be pointers?

pkg/peer/messages.go
```diff
@@ -40,19 +46,6 @@ type DataChannelMessage struct {
...
-type RTCPReceived struct {
-	TrackID string
-	Packets []RTCPPacket
-}
-
-type RTCPPacket struct {
-	Type    RTCPPacketType
-	Content rtcp.Packet
+type KeyFrameRequestReceived struct {
+	common.TrackInfo
```
SimonBrandner, line 50: Is this a PLI, should it be named that way? What about FIRs?

pkg/peer/state/peer_state.go
```diff
@@ -0,0 +1,72 @@
...
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	tracks := make([]*webrtc.TrackRemote, 0, len(p.remoteTracks))
+	for _, track := range p.remoteTracks {
+		tracks = append(tracks, track)
+	}
+
+	return tracks
+}
```
SimonBrandner, line 58: Seems to be unused

SimonBrandner (commented):
I think we're almost there!

pkg/conference/participant/tracker.go
SimonBrandner: > I renamed it to `PublishedTrack` for each function, i.e. `AddTrack` is now `AddPublishedTrack`, and `UpdateTrackMetadata` is now `UpdatePublishedTrackMetadata`. Let me know if it's better (to me it seems a bit too lengthy now, but I don't have a strong opinion on that ).

Thank you! It's way more readable for me now

pkg/peer/state/peer_state.go
daniel-abramov: Yep, I just created it for completeness (since it's in a separate package), but I can remove unused functions. The idea of the `PeerState` was to isolate mutex-protected data in its own structure to prevent users from accessing the fields without locking the mutex. Essentially, we isolate it to ensure that we access things in a thread-safe way.

pkg/peer/messages.go
daniel-abramov: So, since RTCPs [failed a lot](https://github.com/matrix-org/waterfall/issues/87) and resulted in errors, I decided that why don't we just simplify it then by converting all read RTCPs into just some "keyframe request" (so both PLIs and FIRs are just sent as keyframe request messages from the `Peer` so that it's up to the peer to then convert the "keyframe request" in what it thinks makes sense. Currently, all "keyframe requests" are indeed converted into PLIs upon arrival to the concrete peer.

This solved all the RTCP errors that were sent (so it seems like these were indeed FIRs that we tried to forward [might require more work to be properly handled]). Though the freezes in Firefox (and only in Firefox) still happen from time to time, but there are no grey frames anymore 🙂

pkg/peer/state/peer_state.go
SimonBrandner: Yeah, I think we should avoid unused methods...

pkg/conference/participant/track.go
daniel-abramov: Ok, rewritten as a loop. The new version is not identical in terms of semantics, but I think it's ok for now given that ideally, we would want to replace this logic with "send an error to the client in case the given quality is not available".

pkg/common/track_info.go
daniel-abramov: 👌 Moved.

pkg/peer/state/peer_state.go
daniel-abramov: 👌 Removed.

SimonBrandner (approved):
LGTM

Do we wait to resolve the Firefox issue before merging?

@dbkr, do you want to have another look?

daniel-abramov:
> LGTM
> 
> Do we wait to resolve the Firefox issue before merging?
> 
> @dbkr, do you want to have another look?

Yep, the issue with grey frames seems to be a bug, I need to fix it either prior to merging or after that.

PR #88, opened 2022-12-27
Use FIR packets that Pion returned to us when forwarding

daniel-abramov:
Note: this does not seem to fix anything, though it feels more correct since FIRs have an FIR header.

### DIFF
--- a/pkg/peer/peer.go
+++ b/pkg/peer/peer.go
@@ -152,7 +152,7 @@ func (p *Peer[ID]) WriteRTCP(trackID string, packets []RTCPPacket) error {
 			toSend[i] = &rtcp.PictureLossIndication{MediaSSRC: ssrc}
 		case FullIntraRequest:
 			// FIRs are a bit more complicated. They have a sequence number that must be incremented
-			// and the an additional SSRC inside FIR payload. So we rewrite the media SSRC here.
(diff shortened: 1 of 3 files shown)


pkg/peer/peer.go
```diff
@@ -145,11 +145,18 @@ func (p *Peer[ID]) WriteRTCP(trackID string, packets []RTCPPacketType) error {
...
-		switch packet {
+		switch packet.Type {
 		case PictureLossIndicator:
+			// PLIs are trivial, they just have media SSRC and sender SSRC, where the last one
+			// does not seem to matter (based on Pion examples of using these).
 			toSend[i] = &rtcp.PictureLossIndication{MediaSSRC: ssrc}
 		case FullIntraRequest:
-			toSend[i] = &rtcp.FullIntraRequest{MediaSSRC: ssrc}
+			// FIRs are a bit more complicated. They have a sequence number that must be incremented
+			// and the an additional SSRC inside FIR payload. So we rewrite the media SSRC here.
```
SimonBrandner, line 155:
```suggestion
			// and an additional SSRC inside FIR payload. So we rewrite the media SSRC here.
```

SimonBrandner (approved):
Otherwise LGTM

PR #94, opened 2023-01-05
Fix unsound implementation of the RTP monitoring go-routine as well as remove obsolete stuff

### LINKED ISSUE #92: Create the sound implementation of the rtp subscription watchdog
We must monitor the RTP packets on subscriptions and print an error if the subscription does not receive any RTP packets for a long time.

This has 2 purposes:
1. We'll be able to see when the subscription is "stalled" which would explain freezes (in case the freeze occurs, we'll know why).
2. We can use this information to switch simulcast layers on the SFU if we detect that one or another layer is not available, i.e. if a high level is used and becomes unavailable.

daniel-abramov:
This fixes https://github.com/matrix-org/waterfall/issues/92.


pkg/common/watchdog.go
```diff
@@ -0,0 +1,70 @@
...
+		c.channel <- struct{}{}
+		return true
+	}
+
+	return false
+}
+
+// Starts a watchdog that periodically (specified by the configuration) executes a `c.OnTimeout` closure if
+// no messages have been received on a channel for a `c.Timeout`.
+func (c *WatchdogConfig) Start() *WatchdogChannel {
```
SimonBrandner, line 51: The usage of `WatchdogConfig` here feels odd to me, wouldn't a `NewWatchdogChannel(c WatchdogConfig) *WatchdogChannel {}` be better? Or is this the go way to do things?

pkg/common/watchdog.go
```diff
@@ -0,0 +1,70 @@
...
+	// The channel that will be used to inform the watchdog about the reception of a packet.
+	// The watchdog will be stopped once the channel is closed.
+	incoming := make(chan struct{}, UnboundedChannelSize)
+
+	go func() {
+		for {
+			select {
+			case _, ok := <-incoming:
+				if !ok {
+					return
```
SimonBrandner, line 61: Should we set closed to false, log an error or do something here? Just returning feels dangerous?

SimonBrandner (approved):
Otherwise LGTM

pkg/common/watchdog.go
daniel-abramov: We did something similar in `HeartBeat`, so I just followed the same approach, but you're probably right, perhaps it's better to have an explicit `NewWatchdog()` function that accepts the config, let me rework that real quick.

SimonBrandner (approved):

PR #97, opened 2023-01-09
Simulcast packet rewriting

### LINKED ISSUE #89: Grey frame seen by participants who join several seconds after the call is started (video only)
The bug is caused by the incorrect RTP timestamp and sequence numbers set for the newly created tracks. This has been introduced during the simulcast implementation to make on-demand subscriptions (unlike the previous non-simulcast version that created local tracks upon reception of remote tracks and added **the same track** to each participant who tried to subscribe to other participant's feeds).

This goes along with the RTP rewriting story, so must be implemented anyway.

### LINKED ISSUE #91: Do not renegotiate on switching between simulcat layers

### LINKED ISSUE #95: Don't apply simulcast logic to the audio tracks
Currently, we don't really differentiate between audio and video tracks when doing RTP packet forwarding. This does not really harm but causes computations [for audio tracks] that are not necessary (that just consume more CPU than needed).

We should not forward RTP tracks for the audio, but instead, we could create a local track for each audio track and add the same track to each peer who subscribes to the audio feed of others (as we did it prior to the simulcast implementation).

That would spare us some CPU usage and simplify a code at certain places a bit.

daniel-abramov:
Fixes https://github.com/matrix-org/waterfall/issues/91
Fixes https://github.com/matrix-org/waterfall/issues/89
Fixes https://github.com/matrix-org/waterfall/issues/95


pkg/peer/subscription/rtp.go
```diff
@@ -0,0 +1,151 @@
...
+	if p.previouslyForwardedSSRC != p.selectedSSRC {
+		// There is a gap of 1 seqnum to signify to the decoder that the previous frame
+		// was (probably) incomplete. That's why there's a 2 for the seqnum.
+		delta := PacketIdentifiers{1, 2}
+
+		// We make an exception for the very first packet that we're forwarding
+		// so that we start with 0 seqnum and 0 timestamp for the first packet.
+		if p.previouslyForwardedSSRC == 0 {
+			delta = PacketIdentifiers{0, 0}
+		}
```
SimonBrandner, line 70: Can this be an if-else statement?

daniel-abramov: I think I know what you mean :) I.e. if it was Rust I would have written it like this:

```rust
let delta = if p.previouslyForwardedSSRC == 0 {
    PacketIdentifiers::default()
} else {
    PacketIdentifiers::new(1, 2)
};
```

That's what you semantically would love to see, right?

If yes, then I'm also in favor of such initialization, but unfortunately in Go the only way to write it is either like this:

```go
var delta PacketIdentifiers
if p.previouslyForwardedSSRC == 0 {
    delta.sequenceNumber = 0, delta.timestamp = 0
} else {
    delta.sequenceNumber = 2, delta.timestamp = 1
}
```

Or, given that by default Go would initialize things with a default value (which is zero by default), shorter:

```go
var delta PacketIdenfiers
if p.previouslyForwadedSSRC != 0 {
    delta.sequenceNumber = 2, delta.timestamp = 1
}
```

The last one does not look that bad, but it seems like a variation of the current version? (except that the if is "flipped" and the initilization is default / less explicit).

Which one would you prefer?

SimonBrandner:
```go
var delta PacketIdentifiers
if p.previouslyForwardedSSRC == 0 {
    delta.sequenceNumber = 0, delta.timestamp = 0
} else {
    delta.sequenceNumber = 2, delta.timestamp = 1
}
```

^ reads the best from POV or perhaps (think that should work, right?):

```go
var delta PacketIdentifiers
if p.previouslyForwardedSSRC == 0 {
    delta = PacketIdentifiers{0, 0}
} else {
    delta = PacketIdentifiers{2, 1}
}
````

File: pkg/peer/peer.go [hunk 2]
```diff
@@ -99,16 +99,17 @@ func (p *Peer[ID]) SubscribeTo(track common.TrackInfo) (*subscription.Subscripti
...
 
-	// The ssrc that we must use when sending the RTCP packet.
-	// Otherwise the peer won't understand where the packet comes from.
-	ssrc := uint32(track.SSRC())
+	if uint32(track.SSRC()) != info.Simulcast.SSRC {
+		return fmt.Errorf("track ssrc %d doesn't match the one in the info %d", track.SSRC(), info.Simulcast.SSRC)
+	}
 
-	rtcps := []rtcp.Packet{&rtcp.PictureLossIndication{MediaSSRC: ssrc}}
+	p.logger.Debugf("Keyframe request: %s (%s)", info.TrackID, info.Simulcast.Layer)
```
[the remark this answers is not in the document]
SimonBrandner: Yeah, perhaps

pkg/conference/participant/tracker.go
```diff
@@ -200,19 +249,22 @@ func (t *Tracker) Unsubscribe(participantID ID, tracks []TrackID) {
...
+				logrus.Errorf("Bug: subscription without subscriber")
 			}
 		}
 	}
 }
 
 // Processes RTCP packets received on a given track.
-func (t *Tracker) ProcessKeyFrameRequest(info common.TrackInfo) error {
+func (t *Tracker) ProcessKeyFrameRequest(info common.TrackInfo, simulcast common.SimulcastLayer) error {
 	const sendKeyFrameInterval = 500 * time.Millisecond
```
dbkr, line 268: This looks unused now?

pkg/peer/subscription/rtp.go
```diff
@@ -0,0 +1,135 @@
...
+		p.previouslyForwardedSSRC = packet.SSRC
+	} else {
+		// If the incoming timeestamp or sequence number are smaller than the timestamp of the first packet after the switch,
+		// then we're getting the old packet (before switch), which we're not interested in, so we drop it.
+		if incomingIDs.timestamp < p.firstIncomingIDs.timestamp ||
+			incomingIDs.sequenceNumber < p.firstIncomingIDs.sequenceNumber {
+			return nil, fmt.Errorf("Ignoring old packet")
+		}
+
+		delta := incomingIDs.Sub(p.firstIncomingIDs)
```
dbkr, line 78: Bear in mind that these IDs will wrap around, SSRCs quite frequently, so this will end up being negative sooner or later.

pkg/peer/subscription/video.go
```diff
@@ -0,0 +1,145 @@
...
+		nil,
+		logger,
+	}
+
+	// Configure watchdog for the subscription so that we know when we don't receive any new frames.
+	watchdogConfig := common.WatchdogConfig{
+		Timeout: 2 * time.Second,
+		OnTimeout: func() {
+			logger.Warnf("No RTP on subscription for %s (%s)", subscription.info.TrackID, subscription.currentLayer)
+			subscription.requestKeyFrame()
```
dbkr, line 71: Was this going to switch down to whatever layer is still getting packets?

pkg/conference/participant/tracker.go
daniel-abramov: Oh, good catch, thanks! Golang linter unfortunately did not capture that.

pkg/peer/subscription/rtp.go
daniel-abramov: The timestamps and sequence numbers? Won't they be caught by the `if` above?

Sorry, I did not quite get what you mean by wrapping around SSRCs though.

pkg/peer/subscription/video.go
daniel-abramov: Yeah, right, I've specifically created this helper `requestKeyFrame()` which uses `subscription.info` and `subscription.simulcast`, i.e. making sure that we use the actual current layer and not the layer that we used when creating the initial subscription.

pkg/peer/subscription/rtp.go
dbkr: Oh sorry, I meant seqnums not SSRCs.

pkg/peer/subscription/video.go
dbkr: ah, I missed that this was doing this

pkg/peer/peer.go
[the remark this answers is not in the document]
daniel-abramov: This is not a bad idea, though I think if we move it to trace, we probably would never enable it? I.e. ideally key frame requests should not be sent that often, I would say that they are always logged in big quantities, it would signify a bug. During the testing with debug logs I've noticed that they don't really happen that often (primarily during the switch/subscribe/unsubscribe), otherwise, if they were logged too often, there were always some bugs that caused it.

Does it feel like they would not be that bad for the debug logs since we only would use them during the **debugging**? In production, I think we would anyway run the SFU with the default maximum level of `info`. What do you think?

I think tracing is more suitable for things like... if we let's say logged each packet (there would be hundreds of them per second), whereas keyframe requests do not even happen every second.

pkg/peer/subscription/rtp.go
daniel-abramov: Do you by chance know what's the right way to rewrite these values once they wrap? I.e. what do actually browsers expect? I'm not that much concerned about the timestamps since they are `uint32` and so far I've never experienced them wrapping (though I believe it might happen for longer conferences?), but the wrapping of sequence numbers does happen for conferences e.g. conferences that last for half an hour.

Given that the sequence number is just an `uint16`, we reach this rather sooner (2^16=65536). Do you by chance know if the browser would expect a sequence number of `0` after we've sent the one with the sequence number `65535`?

dbkr: Yeah, it just wraps around in the normal way, so 0 comes after 65535. Bear in mind that timestamps don't start from 0 though, so it's not just length of call that's the factor, it's also being unlucky and it choosing a starting TS near the top of the range.

pkg/conference/participant/tracker.go
```diff
@@ -75,6 +76,16 @@ func (t *Tracker) RemoveParticipant(participantID ID) map[string]bool {
...
 
+	// Remove this participant's subscriptions.
+	for _, subscribers := range t.subscribers {
+		for subscriberID, subscription := range subscribers {
+			if subscriberID == participantID {
+				subscription.Unsubscribe()
+				delete(subscribers, subscriberID)
+			}
+		}
+	}
```
SimonBrandner, line 87:
```suggestion
	for _, subscribers := range t.subscribers {
		for subscriberID, subscriber := range subscribers {
			if subscriberID == participantID {
				subscriber.Unsubscribe()
				delete(subscribers, subscriberID)
			}
		}
	}
```

pkg/peer/subscription/audio.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (s *AudioSubscription) readRTCP() {
+	for {
+		if _, _, err := s.sender.ReadRTCP(); err != nil {
+			if errors.Is(err, io.ErrClosedPipe) || errors.Is(err, io.EOF) {
+				return
+			}
+		}
+	}
```
SimonBrandner, line 59: This seems like wasting performence?

daniel-abramov: Hm? What do you mean?

pkg/conference/participant/tracker.go
daniel-abramov: The thing is that the subscriber is **a key** in the map. subscription is **a value**. The actual type that is inside the value is a `Subscription`, that's why I also called the variable `subsscription`.

Perhaps the confusion comes from the fact that `subscribers` is **a map of maps**. `subscribers` is a map where the key is a track ID, while the value is a map that holds the correspondence between the subscriber ID (participant ID) and the subscription, i.e. contains both subscribers and subscriptions.

Or, if it would be simple to understand, `subscription` is semantically a map `map[TrackID]SubscriptionDetails`, where `SubscriptionDetails` is:

```go
type SubscriptionDetails struct {
    subscriber participant.ID
    subscription subscription.Subscription
}
```

Does it make any sense? 🙂

daniel-abramov: In other words: a subscriber is a participant (or rather, its identifier) who is subscribed to something. Subscription is an actual entity (the actual subscription that handles the logic for the subscribed track, processes incoming packets and writes them to the output track).

SimonBrandner: Ah, sorry I am confused

pkg/peer/subscription/audio.go
SimonBrandner: The for loop doesn't seem to be doing anything - why is there?

pkg/peer/subscription/rtp.go
```diff
@@ -0,0 +1,135 @@
...
+	previouslyForwardedSSRC uint32
+	// SSRC that we're using for all packets that we're forwarding.
+	// This is the SSRC that we're sending to the remote peer. Typically,
+	// this is the SSRC of the lowest layer for the simulcast track.
+	outgoingSSRC uint32
+	// The highest identifiers of the forwarded packet,i.e. the IDs of
+	// the **latest** (in terms of timestamp and number) packet. This is not
+	// necessarily the IDs of the **last forwarded packet**, due to packets
+	// that may arrive out-of-order.
+	maxOutgoingIDs PacketIdentifiers
```
SimonBrandner, line 24: While it might not be exactly the last forwarded packet, I still think `lastOutgoingIDs` would be more descriptive and give you a better picture when reading the code for the first time

pkg/peer/subscription/rtp.go
```diff
@@ -0,0 +1,135 @@
...
+		p.firstIncomingIDs = incomingIDs
+		p.firstOutgoingIDs = outgoingIDs
+
+		// Update the SSRC of the previously forwarded packet.
+		p.previouslyForwardedSSRC = packet.SSRC
+	} else {
+		// If the incoming timeestamp or sequence number are smaller than the timestamp of the first packet after the switch,
+		// then we're getting the old packet (before switch), which we're not interested in, so we drop it.
+		if incomingIDs.timestamp < p.firstIncomingIDs.timestamp ||
+			incomingIDs.sequenceNumber < p.firstIncomingIDs.sequenceNumber {
```
SimonBrandner, line 74: Won't this break since they wrap around?

pkg/peer/remote_track.go
```diff
@@ -0,0 +1,92 @@
...
+		rtpBuf := make([]byte, 1400)
+
+		// Call this when this goroutine ends.
+		defer func() {
+			p.sink.Send(PublishedTrackFailed{trackInfo, common.SimulcastLayerNone})
+			p.state.RemoveRemoteTrack(remoteTrack)
+		}()
+
+		for {
+			index, _, readErr := remoteTrack.Read(rtpBuf)
```
SimonBrandner, line 75: Why not use `ReadRTP()`? It always felt a bit simpler to read to me

pkg/peer/messages.go
```diff
@@ -18,16 +18,25 @@ type LeftTheCall struct {
 }
 
 type NewTrackPublished struct {
+	// Information about the track (ID etc).
 	common.TrackInfo
+	// Simulcast configuration (can be `None` for non-simulcast tracks and for audio tracks).
+	Simulcast common.SimulcastLayer
```
SimonBrandner, line 24: I think either `Layer` or `SimulcastLayer` would be more clear on first read

pkg/peer/messages.go
```diff
@@ -18,16 +18,25 @@ type LeftTheCall struct {
...
+	// Output track (if any) that could be used to send data to the peer. Will be `nil` if such
+	// track does not exist, in which case the caller is expected to listen to `RtpPacketReceived`
+	// messages.
+	OutputTrack *webrtc.TrackLocalStaticRTP
 }
 
 type PublishedTrackFailed struct {
 	common.TrackInfo
+	Simulcast common.SimulcastLayer
```
SimonBrandner, line 33: I think either `Layer` or `SimulcastLayer` would be more clear on first read

pkg/peer/messages.go
```diff
@@ -18,16 +18,25 @@ type LeftTheCall struct {
...
 }
 
 type RTPPacketReceived struct {
 	common.TrackInfo
-	Packet *rtp.Packet
+	Simulcast common.SimulcastLayer
```
SimonBrandner, line 38: I think either `Layer` or `SimulcastLayer` would be more clear on first read

pkg/peer/messages.go
```diff
@@ -48,4 +57,5 @@ type DataChannelAvailable struct{}
 
 type KeyFrameRequestReceived struct {
 	common.TrackInfo
+	Simulcast common.SimulcastLayer
```
SimonBrandner, line 60: I think either `Layer` or `SimulcastLayer` would be more clear on first read

SimonBrandner (commented):
Otherwise LGTM

pkg/peer/subscription/audio.go
daniel-abramov: To read RTCPs. I don't know if it's required for audio, but it seems like it's generally something that should be there since reading RTCPs (even if we don't process them) makes Pion process the packet with the interceptors, which is required for NACKs, etc.

Or did you mean something different?

pkg/peer/remote_track.go
daniel-abramov: I agree that `ReadRTP()` is a better choice when we're interested in RTP packets, that's also something that I use in when handling **video** tracks since we are working on a packet level when forwarding and rewriting packet IDs for the video.

In general, I think the only advantage of the `Read()` and `Write() over the `ReadRTP()` is that they use the buffer that you pass to the function, which means that `Read()` does not allocate, since we're reusing the same slice all over again (slight performance advantage). Since we don't really process audio packets in any way and just forward it within a single function, I just decided to use `Read()` / `Write()` here.

However, if you think that it harms the readability of this code, I can change it to the `ReadRTP()` and `WriteRTP()` (I think the performance loss, in this case, might be negligible anyway).

pkg/peer/subscription/rtp.go
SimonBrandner: Personally, I prefer `last` even if slightly wrong... Does anyone else have any opinions on this?

pkg/peer/subscription/audio.go
SimonBrandner: Ah, so `ReadRTCP()` does something even though we don't use the packets it returns?

pkg/peer/subscription/rtp.go
daniel-abramov: The thing is... I might need to use `last` as the last packet that we received when properly handling wrapping, so this name would be occupied 🙂 What about `latest`? Maybe sounds more reasonable than `last` and `max`? (I think I might actually indeed prefer `latest`)

SimonBrandner: `latest` sounds great!

pkg/peer/subscription/audio.go
daniel-abramov: > Ah, so ReadRTCP() does something even though we don't use the packets it returns?

Yeah, right. Pion has a concept of interceptors and uses "default interceptors" if none others are specified. This construction is very [common in Pion examples](https://github.com/pion/webrtc/blob/master/examples/reflect/main.go#L77).

> (are there any RTCP packets for audio tracks even?)

I'm not an expert to be honest, but I assume that things like NACKs (negative acknowledgments) for audio do still exist (i.e. so that the sender indicates packet loss), since we still have sequence numbers (and timestamps) for audio packets (both video and audio are transmitted as RTP streams).

SimonBrandner: Argh, right, this feels weird tbh, Pion. Can we add a comment?

pkg/peer/subscription/rtp.go
daniel-abramov: Done.

pkg/peer/peer.go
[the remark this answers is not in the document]
daniel-abramov: Done.

pkg/peer/subscription/audio.go
daniel-abramov: Added!

pkg/peer/subscription/rtp.go
daniel-abramov: Updated.

SimonBrandner (approved):
LGTM!

PR #99, opened 2023-01-12
Create idiomatic package structure and watchdog

EnricoSchw:
These are only cosmetic changes:
- separated main file from the lib folder
- Set up unit tests
- Watchdog adjusted


pkg/common/watchdog.go
```diff
@@ -68,3 +68,65 @@ func StartWatchdog(c WatchdogConfig) *WatchdogChannel {
...
+	timeout time.Duration
+	// A closure that is called once `Timeout` is reached.
+	onTimeout   func()
+	notify      chan struct{}
+	closeSignal chan struct{}
+}
+
+func NewWatchdog(timeout time.Duration, onTimeout func()) *Watchdog {
+	return &Watchdog{timeout: timeout, onTimeout: onTimeout, notify: make(chan struct{}), closeSignal: make(chan struct{})}
+}
```
daniel-abramov, line 83: Is there a use case for starting a watchdog that does not get started?

I think the previous API surface of the `WatchdogChannel` was somewhat more resilient in this case, i.e. the watchdog is started once created and could be notified and closed after that.

Whereas the new API has an issue that calling `Notify()` before `Start()` would **hang the caller forever**. Also, it allows closing the `Watchdog` that has not been started in the first place.

I think if we want a drop-in replacement, we need to preserve the API surface.

Another concern that I see is that the `notify` channel is unbounded, this is different from our implementation in `main` that uses buffered channels (since we generally don't want to block the caller, the caller merely informs the watchdog so that it continues to run and not call the `OnTimeout`, i.e. one can treat `Watchdog` as a channel that provides a unidirectional heartbeat to the caller, making sure that if no heartbeat messages are received for `timeout`, some action `onTimeout()` is executed).

Also `Start()` can be called multiple times which would make the behavior of the watchdog a bit unpredictable due to separate goroutines racing to receive the `Notify()` message, where only would get it.

pkg/common/watchdog.go
```diff
@@ -68,3 +68,65 @@ func StartWatchdog(c WatchdogConfig) *WatchdogChannel {
...
+
+func (w Watchdog) Start() chan struct{} {
+	terminate := make(chan struct{})
```
daniel-abramov, line 86: Do we need a `terminate` channel here? From what I can see, we don't use (won't use it) and it's only used for the tests.

cmd/sfu/main.go
```diff
@@ -23,6 +23,8 @@ import (
 	"syscall"
 
 	"github.com/matrix-org/waterfall/pkg/config"
+	"github.com/matrix-org/waterfall/pkg/profiling"
```
SimonBrandner, line 26: What's the benefit to moving the main file?

SimonBrandner (commented):
Looks sane otherwise

pkg/common/watchdog_test.go
```diff
@@ -0,0 +1,128 @@
...
+		t.Fatal("Should terminated only after call Close")
+	default:
+		assert.True(t, true)
+	}
+}
+
+func TestWatchdog_Close(t *testing.T) {
+	w := testSetup(t)
+	var terminated chan struct{}
+	terminated = w.Start()
```
daniel-abramov, line 37:
```suggestion
	terminated := w.Start()
```

pkg/common/watchdog_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	w.Close()
+	assert.Empty(t, <-terminated)
+}
+
+func TestWatchdog_Notify(t *testing.T) {
+	w := testSetup(t)
+	w.Start()
+	// Take care: Notify is blocking and this lines running sequential.
+	// The previous implementation was blocking too. I'll leave it like this
```
daniel-abramov, line 53: It was not actually 😉 The previous implementation used buffered channels.

pkg/common/watchdog_test.go
```diff
@@ -0,0 +1,128 @@
...
+	// You can close the watchdog before starting it. I don't know if this is a feature ;-)
+	w := testSetup(t)
+	w.Close()
+	assert.Empty(t, <-w.Start())
+}
+
+func TestWatchdog_Multiple_Start(t *testing.T) {
+	// You can start Watchdog several times.
+	// I think this is a bug, but I didn't want to change the interface.
+	// I would prefer make start private and start the watchdog immediately upon creation to prevent this.
```
daniel-abramov, line 77: I agree that it sounds like a bug, that's why the interface of the previous implementation was designed slightly differently (to now allow calling `Start()` multiple times). Could we probably get the previous interface, please? 🙂 (less possibilities to introduce a bug)

cmd/sfu/main.go
EnricoSchw: - In pkg are public library file. main.go is not a library file.
- You setup in main.go the SFU. You can have several mains with different setups
For example a SFU uses RPC as the signaling layer, another SFU use Data-Channels.

pkg/common/watchdog.go
```diff
@@ -68,3 +68,65 @@ func StartWatchdog(c WatchdogConfig) *WatchdogChannel {
 
 	return &WatchdogChannel{incoming, sync.Mutex{}, false}
 }
+
+type Watchdog struct {
```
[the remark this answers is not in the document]
EnricoSchw: Right, I left it for now but it's dead code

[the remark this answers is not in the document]
SimonBrandner: I'd suggest removing it, we can always use git to get it back

daniel-abramov (commented):
Congrats on the first PR, @EnricoSchw 🙂 

I generally agree with the package changes. For the watchdog: I personally a bit prefer the previous implementation since it was sound yet quite short and [at least to me] easier to comprehend, but I might be biased a bit, so I am by no means against the new implementation, i.e. if it's a more idiomatic way to implement it in Go (common idiom) and as long as @SimonBrandner also ok with it, then I would be glad us using it 🙂 Yet, I left a couple of comments to address certain issues that the current implementation has, i.e. I think we need to write a drop-in replacement preserving the same API surface of the watchdog, otherwise new issues are introduced.

There are also a couple of generic remarks that I would like to make if I may:
1. The previous watchdog implementation had comments, the new one mostly does not, perhaps some comments would not harm (since it's in a `common` module and could be viewed as some sort of a "library component").
2. The previous implementation is still there, i.e. it co-exists with the new one, I think if we use the new implementation, then there is no reason to keep the old one, could you please remove it?
3. You added certain unit tests for the new implementation. Tests are good, I think it's worth it to add tests to our CI, so that it executes along other stages (building, linting).
4. It feels like this PR could be split in two PRs: one changes the package structure, another replaces the watchdog, but no biggie. However, inside a single PR commits should be a bit more structured/atomic (in case you want to merge all commits and not squash all of them into a single one). The problems with current commit history that I spotted are: some commit messages are too long and don't fit in standard commit length (wrapped), there is a merge commit that should not be there (please rebase to master so that this commit is not there), and also the commits are not quite "atomic", i.e. they depend very much on adjacent commits that fix them and contain things that they should not contain, e.g. [this commit](https://github.com/matrix-org/waterfall/pull/99/commits/04115c3397cc67ea489ee32a517f4f57f4ac2696) contains some commented out lines and does too many things (should be split in two). I think this PR generally should have 2 commits: change of the package structure and watchdog implementation.

pkg/common/watchdog.go
```diff
@@ -68,3 +68,41 @@ func StartWatchdog(c WatchdogConfig) *WatchdogChannel {
...
+//			return
+//		} }
+//}()
+//return terminated }
+//done := make(chan interface{}) terminated := doWork(done, nil)
+//go func() {
+//	// Cancel the operation after 1 second. time.Sleep(1 * time.Second) fmt.Println("Canceling doWork goroutine...") close(done)
+//}()
+//<-terminated
+//fmt.Println("Done.")
```
daniel-abramov: ?

pkg/common/watchdog.go
daniel-abramov: Ok, agree!

pkg/common/watchdog.go
EnricoSchw: About the general remarks:

1. I agree, i would write more comments
2. I agree, one implementation has to leave :-)
3. and 4. I will split the PR into two topics. I will rebase them and note your other comments. I will try to include the ci tests in the PR with pkg structure changes.

daniel-abramov: Cool, thanks Enrico!

daniel-abramov: > I would differentiate a little bit about the unbounded channel part. I'm not that happy with my implementation, because this implementation doesn't make the source code much better. Of course, we can use a buffer for the notify channel, but its remains still blocking. Because, when the buffer is reached then sending is blocked again.
>
> This is not a theoretical view of that. 'Notify()' is called in Subscription.WriteRTP(). This method is called whenever a RTP packet arrives. If I say something wrong please correct me, that can be 400 packets per second.

I totally agree with your reasoning. Theoretically, it can and will reach the buffer size at some point (and practically will reach it, though in our case that would mean a bug on our side). The constant that is defined in a `common` package was meant to set the size of the channel to "large enough" that it sort of "never blocks", or if I can re-formulate it: "if it ever reaches this size and blocks then there is a bug somewhere in our code".

Do you know if there is a better way to create a channel that does not have a limited buffer size (i.e. the buffer of which is limited to the size of the RAM, like unbounded channels in Rust)? Alternatively, the only way to ensure that it never blocks (even in theory) would be to do something like `go func() { ch<- struct{}{}}()` or something like this, but it feels very inefficient to spawn a goroutine just for the sending procedure (I know that goroutines are lighter than actual threads, but are they that light to actually be used like this given that we would be doing it hundreds times per second).

We could of course try to use the `select/default` trick though.

EnricoSchw: I once added a benchmark:
```shell
BenchmarkWatchdog_Notify-4               2801523               418.1 ns/op
BenchmarkWatchdogChannel_Notify-4        4755174               243.7 ns/op
```
An RTP packet arrives every 20 ms, there is theoretically enough time :-) Now I think, both implementations are fast enough. 

Unfortunately, my implementation is twice as slow as the original implementation. :-( So we should stay on the old one.

To your question:
I think go routines are fast enough.
But I cannot answer your question in general. I think the blocking behavior of go channels is explicitly intended. I think channels are there to transfer ownership of data from one go routine to the next. And for this process, the sender has to wait until the receiver is ready.
When go routines are started I always ask myself from which go routine is it getting data and to which go routine is data send.

Dockerfile
```diff
@@ -14,13 +14,14 @@ COPY go.sum ./
...
 COPY ./pkg ./pkg
+COPY ./cmd ./cmd
 
 # This runs a bit slower but guarantees that the binary does not rely on
 # the underlying C environment (e.g. as "static" as possible).
 RUN CGO_ENABLED=0 \
     GOOS=linux \
     GOARCH=amd64 \
-    go build -o /waterfall ./pkg
+    go build -o /waterfall ./cmd/sfu
```
dbkr, line 24: Will we also need to fix the nix flake?

EnricoSchw: Oh yes. I forget. I'll do.

PR #109, opened 2023-01-23
Revisited wrap around handling

### LINKED ISSUE #103: Revisit the RTP timestamp and seqNum rewriting logic
There are 3 things that we don't take into account:

- Possible out-of-order delivery of packets ✅ .
- VP8PictureIDs (probably **optional for now**).
- TL0PICIDX (probably **optional for now**).

Also, currently, we don't properly expand the timestamp in case of a rollover: the timestamp gets properly extended, but once the sequence number reaches the sequence number with which the simulcast began, we don't properly expand the timestamp which creates some issues.

daniel-abramov:
Closes https://github.com/matrix-org/waterfall/issues/103.

This should account for all possible rollovers and rollunders and cover with tests all cases that might happen in the future.

Hopefully, this resolves any video freezes.


pkg/peer/subscription/rewriter/rewriter_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{60001, 20002, 2222, 10033, 1000003},   // global counter wrap around (roll over)
+		{60002, 20003, 2222, 10034, 1000004},   // global counter wrap around (after a roll over)
+		{0, 20004, 2222, 15568, 1000005},       // normal packet
+		{15000, 20005, 3333, 15570, 1000006},   // layer switch (SSRC change)
+		{0, 20005, 3333, 570, 1000006},         // woohoo, double wrap around, let's go!!!
+		{5, 20005, 3333, 575, 1000006},         // normal packet
+		{5005, 20005, 3333, 5575, 1000006},     // normal packet
+		{5006, 2000500, 3333, 5576, 2980501},   // normal packet
+		{5006, 4294967295, 3333, 5576, 980000}, // global ts wrap around (roll over)
+		{5006, 0, 3333, 5576, 980001},          // and local ts wrap around as well (roll over)
```
SimonBrandner, line 36: Can we have a case of both rollover and SSRC change (unless it's already there and I looked it over)?

SimonBrandner (approved):

SimonBrandner:
(@dbkr might also want to take a look since I am not confident about checking the expands stuff)

pkg/peer/subscription/rewriter/rewriter_test.go
daniel-abramov: We do have this case, but not caused by a single packet 🙂 I think the only possible case this might happen is when e.g. our seqNum is 65535 and we get the SSRC change if I'm getting you right?

dbkr (approved):
Yeah, I think this makes sense

pkg/peer/subscription/rewriter/rewriter_test.go
SimonBrandner: Ah, yeah, that is quite rare

daniel-abramov: Right, it's quite rare, though a valid point! So I added a test for this case as well as you suggested.

PR #110, opened 2023-01-24
Run each subscription's packet processing in its own worker

### LINKED ISSUE #108: Going beyond a single core for subscription handling
Currently, the conferences will become quite unstable if we have more than 100 subscriptions. This happens since each conference has its own goroutine (or thread) which handles all events happening in the session, including handling of RTP packets received from the publishers and writing them to the subscribers. Essentially, this means that the main conference loop handles potentially [short] CPU-bound operations that at a certain point (when there are more than 100 down tracks / 100 subscriptions) start to lag behind due to this thing scaling linearly to the number of subscriptions. This means that if we have a machine with multiple cores, the SFU will scale horizontally well (each conference has its own goroutine), but not much vertically.

Since writing packets to tracks is an expensive thing by itself, we must utilize a more efficient work-stealing thread pool (a pool of workers to be more precise) to parallelize the handling of packets for independent subscriptions.

Related articles:
- LiveKit's way to [scale beyond a single core](https://medium.com/livekit/going-beyond-a-single-core-4a464d20d17a).
- [Signal's article on scaling the calls](https://signal.org/blog/how-to-build-encrypted-group-calls/).

daniel-abramov:
I've realised that actually that generic helper that I created for the subscriptions to track missing packets (watchdog) is very close in semantics to the concept of a single worker. In fact, the watchdog is just a special case of a worker where the task is an empty `{}` and the task handler is an empty function. So I decided to generalize it a bit and that almost automatically allowed me to transform watchdog into a subscription worker which handles incoming packets. So it was a fairly elegant transformation that did not touch many parts of the code 🙂 

So now, each call to the `Subscription.WriteRTP()` does not block, but merely sends the packet to the worker that then handles the packets.

I've also made it non-blocking since if the channel blocks in this case, it means that the worker can't keep up with the load and so packet drops are inevitable anyway.

Fixes https://github.com/matrix-org/waterfall/issues/108.

### DIFF
--- a/pkg/common/worker.go
+++ b/pkg/common/worker.go
@@ -1,8 +1,15 @@
 package common
 
 import (
+	"errors"
 	"sync"
 	"time"
+)
+
+// Errors that may occur when sending tasks to a worker.
+var (
+	ErrWorkerClosed  = errors.New("worker is closed")
+	ErrWorkerTooBusy = errors.New("worker is already overloaded")
 )
 
 // Configuration for the worker.
@@ -36,16 +43,24 @@ func (c *Worker[T]) Stop() {
 
 // Send a task to the worker. Returns `true` if the task
 // has been sent, `false` if the channel is already closed.
-func (c *Worker[T]) Send(task T) bool {
+func (c *Worker[T]) Send(task T) error {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
+	// If the channel is not closed, check if we can send the task.
 	if !c.closed {
-		c.channel <- task
-		return true
+		// We don't want to block here since it's the whole point of this
+		// component (that the CPU bound tasks are handled by the worker).
+		select {
+		case c.channel <- task:
+			return nil
+		default:
+			return ErrWorkerTooBusy
+		}
 	}
 
-	return false
+	// Otherwise, the channel is closed.
+	return ErrWorkerClosed
 }
 
 // Starts a worker that periodically (specified by the configuration) executes a `c.OnTimeout` closure if
--- a/pkg/common/worker.go
+++ b/pkg/common/worker.go
@@ -14,6 +14,8 @@ var (
 
 // Configuration for the worker.
 type WorkerConfig[T any] struct {
+	// The size of the bounded channel.
+	ChannelSize int
 	// Timeout after which `OnTimeout` is called.
 	Timeout time.Duration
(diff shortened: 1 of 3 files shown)


pkg/common/worker.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// If the channel is not closed, check if we can send the task.
+	if !c.closed {
+		// We don't want to block here since it's the whole point of this
+		// component (that the CPU bound tasks are handled by the worker).
+		select {
+		case c.channel <- task:
+			return nil
+		default:
+			return ErrWorkerTooBusy
```
dbkr, line 60: Will this risk dropping the most recent packets and therefore the SFU ending up permanently behind? We presumably still need to keep this a last resort, but probably need the worker OnTask function to drop the oldest packets in the queue in order to catch up?

daniel-abramov: Yes, you're right, the approach that you suggest is also something that I think is a better choice. The only reason I did not implement it is that I thought that it might be a bit complicated (especially with packet rewriting), so it probably requires a bit more thinking, so I thought a "smarter" way (the way you suggested) to drop older packets is better to implement once we land the support of keyframe detection and switching on key frames since it would allow us to implement the logic (drop all non-key-frame packets from the queue when the queue gets full too quickly).

One thing to note is that if we really return `ErrWorkerTooBusy`, then it would mean that the conference is quite overloaded already and the SFU can't keep up with the new subscriptions. So ideally it should not happen at all for conferences, I may assume that we may get this error only when there are too many things going on in the SFU.

SimonBrandner (approved):

pkg/common/worker_test.go
```diff
@@ -0,0 +1,23 @@
...
+func BenchmarkWorker(b *testing.B) {
+	workerConfig := common.WorkerConfig[struct{}]{
+		Timeout:   2 * time.Second,
+		OnTimeout: func() {},
+		OnTask:    func(struct{}) {},
+	}
+	w := common.StartWorker(workerConfig)
+
+	for n := 0; n < b.N; n++ {
+		w.Send(struct{}{})
```
EnricoSchw, line 19: This always returns an error, because we have to add an buffer for the channel in the WorkerConfig.

daniel-abramov: Oh, good catch! 👍 Thanks!

EnricoSchw (approved):

PR #111, opened 2023-01-25
Allow configuration of a custom IP for the SFU (public IP)

daniel-abramov:


fkwp:
would it be possible to configure an array of IP addresses (IPv4 and IPv6?) or is that configures IP on top of the others?

daniel-abramov:
> would it be possible to configure an array of IP addresses (IPv4 and IPv6?) or is that configures IP on top of the others?

Yep, right, the array of IP addresses is very doable. Should I make it to support it? (it's a trivial thing 🙂).

Basically, the public IP address(es) that we give to Pion are the IP addresses that Pion will use as host candidates instead of using private IP addresses.

However, one can specify that Pion does not replace private IP addresses and instead adds them to the list of server reflexive candidates, but I'm not sure it's desired.

config.yaml.sample
```diff
@@ -1,8 +1,12 @@
 matrix:
-  homeserverUrl: "http://localhost:8008"
-  userId: "@sfu:shadowfax"
-  accessToken: "..."
+  homeserverUrl: "http://localhost:8008" # The URL of the home server
+  userId: "@sfu:shadowfax"               # The MXID
+  accessToken: "..."                     # Access token
```
SimonBrandner, line 4:
```suggestion
  userId: "@sfu:shadowfax"               # The MXID of the SFU user
  accessToken: "..."                     # Access token of the SFU user
```

pkg/webrtc_ext/config.go
```diff
@@ -0,0 +1,9 @@
+package webrtc_ext
+
+// Configuration of the WebRTC API for the SFU.
+type Config struct {
+	// Enable simulcast extension.
+	EnableSimulcast bool `yaml:"simulcast"`
+	// Pulibc IP address of the SFU.
+	PublicIP string `yaml:"ip"`
```
SimonBrandner, line 8: As @fkwp suggested, it would be great if this could be an array

pkg/webrtc_ext/webrtc_api.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Creates Pion's WebRTC API that has all required extensions configured (such as simulcast).
+func createWebRTCAPI(config Config) (*webrtc.API, error) {
+	mediaEngine := &webrtc.MediaEngine{}
+	if err := mediaEngine.RegisterDefaultCodecs(); err != nil {
+		return nil, fmt.Errorf("failed to register default codecs: %w", err)
+	}
+
+	// Enable extension headers needed for simulcast (if enabled).
+	if config.EnableSimulcast {
```
SimonBrandner, line 18: Is the current SFU code actually prepared to work with simulcast disabled?

SimonBrandner (commented):
Looking good

pkg/webrtc_ext/config.go
daniel-abramov: Agree, makes sense!

pkg/webrtc_ext/webrtc_api.go
daniel-abramov: Yep, it should work without the simulcast, albeit with a little overhead of RTP rewriting that would not be necessary (won't harm though). Or at least it worked last time I tried it, I would probably a non-Simulcast EC client to test. We currently have a single place where we check if we need to use layers and it has that `if` block that checks if we have an audio type for a track **or** no layers supplied, then we just use `SimulcastLayerNone` for everyone and the rest works identical.

SimonBrandner (approved):

PR #113, opened 2023-01-26
Fix simulcast screen sharing resolution being low all the time

daniel-abramov:


pkg/conference/participant/track_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	cases := []struct {
+		availableLayers             []common.SimulcastLayer
+		fullWidth, fullHeight       int
+		desiredWidth, desiredHeight int
+		expectedOptimalLayer        common.SimulcastLayer
+	}{
+		{layers(low, mid, high), 1728, 1056, 878, 799, mid},   // Screen sharing (Dave's case).
+		{layers(low, mid, high), 1920, 1080, 320, 240, low},   // max=1080p, desired=240p, result=low.
+		{layers(low, mid, high), 1920, 1080, 1900, 1000, mid}, // max=1080p, desired=1080pish, result=mid.
```
SimonBrandner, line 28: This is actually a case where I would still expect the full res, so that we can read text fine

daniel-abramov: But then we would need to change the current behavior which of `receivedResolution <= desiredResolution` and turn it into a different logic where the `receivedResolution` may be larger as well as smaller than the `desiredResolution`, i.e. that we choose the **closest** one. So far the behavior has been that the `desiredResolution` that the client sends defines the upper limit for the resolution.

Do we want to change this behavior right now? - I'm just asking as it looks like we only want this behavior for the screen sharing (because the small/medium-sized text won't be visible unless it's a high resolution) and I'm not sure it would be worth the effort. My ideal suggestion would be for us to support temporal layers for screen sharing (Jitsi and LiveKit do it) since for the screen sharing the low quality does not make any sense (low or medium resolution are oftentimes barely readable, so they don't have much of a user when the screen sharing is used). That way we'll send let's say 15 FPS when we have enough bandwidth and lower the FPS (without sacrificing the quality) if we don't have enough bandwidth. I experienced this yesterday when observing the screen sharing from Florian: it varied between 3 and 10 FPS for me, but the quality remained the same (so I could read and follow regardless of the connection issues).

What do you think?

SimonBrandner: Let's leave it for now and discuss further if necessary

SimonBrandner (approved):

PR #114, opened 2023-01-26
Do not hangup in `onConnectionStateChanged()`

SimonBrandner:


pkg/peer/webrtc_callbacks.go
```diff
@@ -76,8 +75,6 @@ func (p *Peer[ID]) onConnectionStateChanged(state webrtc.PeerConnectionState) {
 	p.logger.Infof("Connection state changed: %v", state)
 
 	switch state {
-	case webrtc.PeerConnectionStateFailed, webrtc.PeerConnectionStateDisconnected, webrtc.PeerConnectionStateClosed:
-		p.sink.Send(LeftTheCall{event.CallHangupUserHangup})
```
dbkr, line 80: We should still do it in failed (and presumably closed) so I'd just remove `webrtc.PeerConnectionStateDisconnected` from the middle.

dbkr (approved):
Thanks, lgtm!

daniel-abramov:
Sorry @SimonBrandner, I'm missing a bit of a context. Why don't we want to hang up on a disconnect? (I could assume that it's because of the reconnect that would be expected on the client side?)

SimonBrandner:
> Sorry @SimonBrandner, I'm missing a bit of a context. Why don't we want to hang up on a disconnect? (I could assume that it's because of the reconnect that would be expected on the client side?)

In the js-sdk we never send a hangup event in the case where `connectionstatechange` is fired, so we're assuming that we fire it incorrectly here. From trying to understand the documentation we gathered that we might be able to recover from `disconnected` and we've removed it here.

Further context: https://matrix.to/#/!dXHEZfLxIQzAELxzQd:matrix.org/$JJmHD9GTUZjmVLERRHZhDpoSXPMXqgcAeiVzpYKzB0Q?via=matrix.org&via=element.io&via=one.ems.host (and messages below)

daniel-abramov:
Makes sense, thanks!

PR #119, opened 2023-01-31
Make it possible to enable trace logging

dbkr:
Also fail on unknown log level strings in the config

### DIFF
--- a/cmd/sfu/main.go
+++ b/cmd/sfu/main.go
@@ -72,6 +72,8 @@ func main() {
 	}
 
 	switch config.LogLevel {
+	case "trace":
+		logrus.SetLevel(logrus.TraceLevel)
 	case "debug":
 		logrus.SetLevel(logrus.DebugLevel)
 	case "info":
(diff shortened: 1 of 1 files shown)


cmd/sfu/main.go
```diff
@@ -72,8 +72,11 @@ func main() {
 	}
 
 	switch config.LogLevel {
+	case "trace":
+		logrus.SetLevel(logrus.TraceLevel)
 	case "debug":
 		logrus.SetLevel(logrus.DebugLevel)
+	case "": // default to info level if unset
 	case "info":
```
daniel-abramov, line 80:
```suggestion
	case "info", "": // default to info level if unset
```

cmd/sfu/main.go
```diff
@@ -85,7 +88,7 @@ func main() {
 	case "panic":
 		logrus.SetLevel(logrus.PanicLevel)
 	default:
-		logrus.SetLevel(logrus.InfoLevel)
+        logrus.Fatalf("unrecognised log level: %s", config.LogLevel)
```
daniel-abramov, line 91: It probably would fail due to the formatting :)

dbkr: ah yep, that's what I get for editing in vim rather than opening up an IDE.

daniel-abramov (approved):

Issue #117, opened 2023-01-30
Make each publisher send RTP packets directly to its subscribers


daniel-abramov:
Currently, each publisher sends incoming RTP packets to the conference. The conference distributes the packets to the subscriptions. The conference does not really do any CPU-bound conferences and merely relays the packets to the subscribers' workers.

However, as it appeared during our demo tests with many participants, this way of sending frames is not very efficient as the frame processing takes over 90% of the whole conference time (if not more). It's not yet clear why (see https://github.com/matrix-org/waterfall/issues/116), but it seems like a good idea to make publishers write to subscribers directly. This would involve sharding the conference state to make subscribers a shared state that is shared between several go-routines / threads.

PR #123, opened 2023-02-02
Performance optimisations for the FOSDEM

### LINKED ISSUE #116: Profile the SFU to understand why packet handling does not happen fast enough
Currently, when we have a conference with 15+ participants, we quickly reach the point where the conference is overloaded with messages and we can't keep up with handling all incoming packets and our queues get full.

It's not clear why it takes so long to process them.

daniel-abramov:
This fixes issues found during https://github.com/matrix-org/waterfall/issues/116

Closes https://github.com/matrix-org/waterfall/issues/116.

Essentially, it does 2 things:

- Moves out the sending of the outbound To-Device messages to a separate worker (so that we never block the main loop of the conference like before when we had to wait for the I/O of writing/sending the To-Device request).
- Previously when we got a peer message, we initially looked up the participant in the map that originated this message. This was and is very convenient, but it's not required to handle all messages, i.e. the message about the new incoming RTP packet does not require access to the participant (as we simply forward this frame to the subscriptions). This optimisation might be reverted once https://github.com/matrix-org/waterfall/issues/117 lands.


pkg/conference/matrix_worker.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (w *matrixWorker) sendSignalingMessage(recipient signaling.MatrixRecipient, content interface{}) {
+	msg := signaling.MatrixMessage{
+		Recipient: recipient,
+		Message:   content,
+	}
+
+	if err := w.worker.Send(msg); err != nil {
+		logrus.Panic("Really bad, dropping matrix message since the matrix queue is full! Home server down?")
```
dbkr, line 44: Is it bad enough to panic the whole SFU though?

daniel-abramov: Ah, you're right, I assumed `Panic` would just print a critical error or something (`Fatal` is documented to exit the process, so I thought that `Panic` would be `Critical` or something, but then checked the source code an found out that it panics). Will fix it right away!

daniel-abramov: Fixed.

dbkr (approved):

PR #127, opened 2023-02-09
Subscription simplification

daniel-abramov:
In order to keep PRs smaller, this is the change that simplifies the way subscriptions are handled a little bit and makes it easier to manage them which is required for the https://github.com/matrix-org/waterfall/issues/117


pkg/conference/participant/tracker.go
```diff
@@ -123,19 +118,10 @@ func (t *Tracker) AddPublishedTrack(
...
-}
-
-// Iterates over published tracks and calls a closure upon each track.
-func (t *Tracker) ForEachPublishedTrack(fn func(TrackID, PublishedTrack)) {
-	for id, track := range t.publishedTracks {
-		fn(id, track)
+// Iterates over published tracks and calls a closure upon each track info.
+func (t *Tracker) ForEachPublishedTrackInfo(fn func(ID, webrtc_ext.TrackInfo)) {
+	for _, track := range t.publishedTracks {
+		fn(track.Owner, track.Info)
```
dbkr, line 124: I realise this isn't related to this change, but is a function like this gaining much convenience / legibility? It seems like it's about as complex as the for loop but losing the ability to use loop control statements. Not important, just an observation.

dbkr (approved):
Thanks — this feels like a better size for a PR and much easier to review.

pkg/conference/participant/tracker.go
daniel-abramov: I see what you mean. I think that if I just had a function `GetPublishedTrackInfo()` that returns an array of these values and allows the caller to just iterate and call their function directly, it would be a simpler way to do things, right? - If that's what you meant, then I totally agree with you 👍 I probably should change it.

The reason why I initially introduced this is that I did not want to return `t.publishedTracks` map to the caller as it would allow the caller to mutate the tracks inside it, which would essentially mean that `publishedTracks` are not really private, but public since the caller can change it. I wanted to forbid it so that `Tracker` can keep the invariants related to `publishedTracks` intact.

However, one could easily achieve the same logic by just creating an array of the capacity of `t.publishedTrack` and just returning that array (with a copy of the required data) to the caller, so probably the `fn` thing is not worth it.

PR #134, opened 2023-02-15
Performant RTP Publisher

### LINKED ISSUE #117: Make each publisher send RTP packets directly to its subscribers

daniel-abramov:
_(per commit review is recommended to follow the changes easier; only one of the commits [the latest one] is relatively large due to moving many things around)_

A rough idea of the current conference is that the conference owns the state of the conference and was the only entity that modified and mutated the state of the conference. That was convenient as there is no need for any locks and the access to the data was sequentialised (see the sketch: https://github.com/matrix-org/waterfall/pull/52). This worked under the assumption that processing a single message takes a very short time and does not really block the conference (i.e. handling all incoming events from the peers, Matrix signaling, etc).

At some, it was clear that certain functions were blocking, while others require more time to get processed than expected, so all blocking operations have been gradually moved to their separate workers/go-routines where the conference loop still received the messages, but offloaded the expensive processing of some of them to these workers while maintaining the state of the conference. That was implemented for e.g. processing the incoming RTP packets from the peers where each subscription had its own worker to process messages. Then the same was repeated for outgoing matrix signaling messages that also were moved into their queue.

While this helped, this did not really solve all the problems since the incoming RTP packets from peers were still sent to the conference (i.e. shared the same message queue) and while the RTP packets **were not processed** by the conference (the conference just read the packet from the queue, identified which subscription it belonged and dispatched it to the right worker), it still was not optimal as any slight lag or hiccup in the processing of any of the events would potentially delay the dispatching of all incoming RTP packets to their workers which means that a seldom lag could cost us several seconds (or even minutes) of freezes.

Sending RTP packets to the conference was not the best idea and it looked like it does not really belong there given that these packets don’t even mutate the conference state and are only relevant for their particular subscribers, so it was clear that the RTP packet handling must be moved out of the main conference queue processing.

This PR changes this by introducing a new logic where each peer only sends a message to the conference informing it about a new available track (or a simulcast quality). The conference then processes this message by creating **a published track** (that in turn creates **a publisher** that runs in its own go-routine). The publisher then reads packets from the incoming track and sends them to the subscriber workers (as the conference did before). This means that the RTP processing is now completely independent from the conference and runs in parallel with another packet handling. The published track (and the publisher) only share the list of available subscriptions with the conference (common shared data structure) the access to which is synchronized, which means that the only moment we currently lock/synchronize the access to the subscriptions is when the state of the subscription changes (layer changed, subscription starts/stops, etc).

This means that the existing publishers and subscribers are not really affected by the main conference loop much and once started, will continue to run efficiently.

Another thing that it allowed us to make is to regulate the size of the queue for the incoming RTP packets that each subscription gets. Since now each published track’s message about the incoming RTP is not sent on a ‘global’ conference queue, we have more control when it comes to the implementation of the backpressure. Currently, each publisher’s worker loop essentially looks like this:

```rust
packet = track.read()
if packet.error() {
    inform_conference_that_the_track_is_gone()
    return
}

for subscription in subscriptions {
    queue_full = subscription.enqueue(packet)
    if queue_full {
        // the packet was not delivered, so we drop this packet
        // log it, todo: we probably want to take the old packets from the queue here
    }
}
```

So now it looks almost exactly the same as the processing of publishers in LiveKit: https://blog.livekit.io/going-beyond-a-single-core-4a464d20d17a/ (except that we went for the approach where we spawn a go-routine per publisher instead of creating a work-pool with a limited amount of workers, as it seems like the last one is a bit more complicated, but does not bring **that much of a performance advantage**).

---

Fixes https://github.com/matrix-org/waterfall/issues/117
Relates to https://github.com/matrix-org/waterfall/issues/120 (but only for RTP packets)


pkg/conference/track/internal.go
```diff
@@ -0,0 +1,103 @@
...
+	requestKeyFrame func(track *webrtc.TrackRemote) error
+}
+
+type audioTrack struct {
+	// The sink of this audio track packets.
+	outputTrack *webrtc.TrackLocalStaticRTP
+}
+
+type videoTrack struct {
+	// Publisher's of each video layer.
```
dbkr, line 21:
```suggestion
	// Publishers of each video layer.
```

pkg/conference/participant/tracker.go
```diff
@@ -3,26 +3,34 @@ package participant
 import (
 	"fmt"
 
-	"github.com/matrix-org/waterfall/pkg/conference/subscription"
+	pub "github.com/matrix-org/waterfall/pkg/conference/track"
```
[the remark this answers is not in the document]
daniel-abramov: Yeah, it is, I think the reason I added it is that without the alias it would be `track.XXX` while `track` is also a very common variable name in the code, so it clashes very often and gives a confusing error. Whereas with `pub` it's just `pub.PublishedTrack` which does not look too bad.

But I could remove the alias and rather rename the variables from `track` to `trackRemote`, `publishedTrack`, etc. What would you prefer?

pkg/conference/participant/tracker.go
```diff
@@ -3,26 +3,34 @@ package participant
...
+
 // Tracks participants and their corresponding tracks.
 // These are grouped together as the field in this structure must be kept synchronized.
 type Tracker struct {
 	participants    map[ID]*Participant
-	publishedTracks map[TrackID]*PublishedTrack
+	publishedTracks map[pub.TrackID]*pub.PublishedTrack[ID]
+
+	publishedTrackStopped chan<- TrackStoppedMessage
+	conferenceEnded       <-chan struct{}
```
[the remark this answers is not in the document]
daniel-abramov: Is seems like, i.e. I think it's very common to see `chan struct{}` as a simple channel to signalize something (usually the "doneness" of something), @EnricoSchw, correct me if I'm wrong.

Such a channel is normally not used to send messages, but rather to signalize that something is finished or done, i.e. the only operation with this channel from the sender's point of view is `close()`, whereas listeners just combine their actions with `<-done` in a `select` statement.

pkg/conference/participant/tracker.go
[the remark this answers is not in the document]
dbkr: Possibly, yes, this seems like a symptom of having a namespace clash in the first place. It might be better to try & avoid the namespace clash by not calling things `track`, as you suggest.

pkg/conference/participant/tracker.go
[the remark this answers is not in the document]
dbkr: OK: I'd seen `chan bool` used for this but I hadn't seen this before. As long as we're consistent within the codebase I think that's fine.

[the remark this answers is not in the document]
daniel-abramov: Hm, I have not seen `chan bool` (I believe using a type, e.g. `bool` might signify that there is a meaning, i.e. that there are values `true` and `false` that may mean something, whereas the unit type `struct {}` can't convey the information). Typically, no messages are sent on a `struct {}` channel. The channel is only used once to close it (the readers only use it for reading to combine with other parts in the `select` statement).

Go's `context` [also seems to be using this pattern](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/context/context.go;l=103) (also the pipelines patterns that are referenced in a comment).

pkg/conference/participant/tracker.go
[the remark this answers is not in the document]
daniel-abramov: Yeah, agree. Originally I used the `pub` alias for `track` package since the `track` is quite a simple and a likely name that a variable can have, so the conflicts are not rare. Alternatively one could rename the package from `track` to something like `published_track` (which is more descriptive for a module), but such package names is almost considered an antipattern for Go and since regardless of the path, only the latest package name is used as a "default" alias, clashes are not very uncommon in Go unfortunately (while using aliases is also not supper common).

Anyway, I've removed the `pub` alias (in fairness "pub" is not a very descriptive alias either) and renamed the variables inside a function. Ptal 🙂

pkg/conference/participant/tracker.go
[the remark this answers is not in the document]
dbkr: ah ok: `chan struct{}` does seem better for that purpose then.

pkg/conference/participant/tracker.go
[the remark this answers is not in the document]
dbkr: Thanks!

dbkr (approved):

pkg/conference/participant/tracker.go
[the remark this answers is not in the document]
EnricoSchw: @daniel-abramov @dbkr `chan struct{}` ist a common pattern for signal channels because an empty struct consumes zero bytes.

pkg/conference/publisher/publisher.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+	"sync"
+
+	"github.com/pion/rtp"
+	"github.com/sirupsen/logrus"
+)
+
+var ErrSubscriptionExists = errors.New("subscription already exists")
+
+type Subscription interface {
```
EnricoSchw, line 13: this can be private like:

```
type subscription interface
```

pkg/conference/publisher/track.go
```diff
@@ -0,0 +1,18 @@
...
+type RemoteTrack struct {
+	// The underlying `webrtc.TrackRemote`.
+	Track *webrtc.TrackRemote
+}
+
+// Implement the `Track` interface for the `webrtc.TrackRemote`.
+func (t *RemoteTrack) ReadPacket() (*rtp.Packet, error) {
+	packet, _, err := t.Track.ReadRTP()
+	return packet, err
+}
```
EnricoSchw, line 18: The source code would be simple if your `publisher.Track` interface directly offered `ReadRTP() (*rtp.Packet, interceptor.Attributes, error)`.

Is the reason because you want hide the `interceptor.Attributes`?

pkg/conference/publisher/publisher.go
daniel-abramov: Will it work? I.e. the `Publisher.AddSubscription()` function and `Publisher.RemoveSubscription()` are both public functions that accept `subscription Subscription`.

pkg/conference/publisher/publisher.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Write the packet to all subscribers.
+	for subscription := range p.subscriptions {
+		if err := subscription.WriteRTP(*packet); err != nil {
+			p.logger.Warnf("packet dropped on the subscription: %s", err)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
daniel-abramov: Thanks!

> An when you move the method: [extractRemoteTrack](https://github.com/matrix-org/waterfall/blob/5e53c5331bc759960511af4398eef8f2a6f6ee98/pkg/conference/track/keyframe.go#L33) from the track pkg to the publisher pkg you can make the Track interface private.

But wouldn't it defeat the whole purpose of having an interface? 🙂 I.e. currently the `Publisher` is quite generic, so we can use it with any track that provides capabilities of reading from and to any subscription that provides the capability of writing to it. Making `Track` private would mean that we would be bound to a specific type of a `RemoteTrack` as the track that could be used with the `Publisher` (which is not bad in general given that we don't use any other types of track **at the moment**, but would arguably be a less generic code, i.e. harder to write a unit test for if the type is fixed, while with interfaces we could provide mock objects easier)

pkg/conference/publisher/track.go
daniel-abramov: That's right 🙂 It seems like interceptor logic is a Pion's internal that [in general] has nothing to do with the publishing logic, so keeping `Track` as a more abstract interface makes it less dependent on internals that we don't necessarily need.

pkg/conference/publisher/publisher.go
EnricoSchw: Yes, because a struct doesn’t have to implement an interface in Go. Go use structural typing (duck typing). This follows the Interface segregation principle. In Go interfaces should be small as possible and they not describe what you offer they describe what you need.

pkg/conference/publisher/publisher.go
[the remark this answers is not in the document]
EnricoSchw: That's an important key question you're asking. Go is not an object oriented language such as Java or C++. I would say Go is structure orientated language. In Java you hide the implementation behind an interface. So you can offer different implementations. In Go you try to base on minimal external structures and have you have only one implementation.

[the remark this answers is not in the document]
EnricoSchw: Maybe an interesting fact that underlines this. In Go there is no inheritance concept. Behavior reuse can only be achieved through composition.

pkg/conference/participant/tracker.go
```diff
@@ -85,151 +90,98 @@ func (t *Tracker) RemoveParticipant(participantID ID) map[string]bool {
...
+	// Wait for the track to complete and inform the conference about it.
+	go func() {
+		// Wait for the track to complete.
+		<-published.Done()
+
+		// Inform the conference that the track is gone. Or stop the go-routine if the conference stopped.
+		select {
+		case t.publishedTrackStopped <- TrackStoppedMessage{remoteTrack.ID(), participantID}:
+		case <-t.conferenceEnded:
+		}
```
EnricoSchw, line 130: Select chooses only one case at random if both are ready. The secound case is to to prevent this goroutine from blocking indefinitely? Then the channel `t.publishedTrackStopped` should never be closed, or we'll panic at some point. on the other hand an unclosed channel cannot be cleaned up by the garbage collector.

EnricoSchw: If we were sure thats `t.conferenceEnded` is closed before `t.publishedTrackStopped` is closed and cleaned up, we could doing something like this:
```go
select {
   case <-t.conferenceEnded: 
      return
   default: 
      t.publishedTrackStopped <- TrackStoppedMessage{remoteTrack.ID(), participantID}:	
}
```

EnricoSchw: A condition thats this not blocking endless is thats all messages from `t.publishedTrackStopped` are read before `close(t.publishedTrackStopped)`

EnricoSchw: We often have that a send channel depends on another open state (here is conference still existsing)

What do you think about this? Has this side effects?

```go
trackIsDone = c.tracker.AddPublishedTrack(sender, msg.RemoteTrack, trackMetadata)
go func() {
   select {
      case <- trackIsDone: 
           c.tracker.RemovePublishedTrack(id track.TrackID)
      case <- t.conferenceEnded:
          // stop track
   }
}
```

pkg/conference/publisher/publisher.go
daniel-abramov: Right, yet I still don't think making an interface private would work. `Subscription` interface is used in a publicly-exposed function which means that the types that it uses must be public as well, otherwise we're using unexported/private types in the exported/public functions.

I quickly tried to change the type to private and indeed it stopped compiling.

pkg/conference/participant/tracker.go
EnricoSchw: > Are you sure about that? - AFAIK, the channels are freed by the garbage collector once they are not used.

No. I was wrong with that. I was talking about buffered channels. But here as well, when the channel is not
referenced, the values in the channel buffer are also not referenced. An thy get cleaned from the gcc  (Ian Lance Taylor, Dave Cheney, I think we can trust them :D)

So we have not to close channels, from this point of few.

pkg/conference/publisher/publisher.go
EnricoSchw: I changed this in this way, and it compiling.

```go
type subscription interface {
	// WriteRTP **must not** block (wait on I/O).
	WriteRTP(packet rtp.Packet) error
}
```

daniel-abramov: Ah, right, it compiles as well, I confused it with the `gopls` warning that `subscription` would shadow a variable with the same name.

Interesting! But do the callers to `AddSubscription`, `RemoveSubscription` know about the private type? I.e. if I were to use the `publisher` package on its own, would I as a caller know the requirements of the interface despite it being private?

(I'm now wondering what's the difference between public and private interfaces then, is it only different in a way that public interfaces could be created in an external code that uses the package? i.e. in case we were **to return** the `subscription`, we would need to make it public)

pkg/conference/publisher/publisher.go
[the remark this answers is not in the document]
EnricoSchw: There are two purpose for interfaces: 
1. Describe conditions your code need. An interface as method parameters.
2. Hide implementations. For example a method signature has an interface as return value.

The second purpose leads more or less to inheritance. Inheritance leads often to strong coupling. This is the reason in Go no inheritance concept is existing. So, People do the conclusion: In Go you should avoid the second purpose of interfaces. (Maybe someone else has a more accessible explanation than I do right now :D)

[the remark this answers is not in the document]
EnricoSchw: There is a role called: `accept interfaces return structs`. Maybe if you google this its helps more as my bad explanations.

EnricoSchw (approved):

pkg/conference/publisher/publisher.go
EnricoSchw: I think we're having the same conversation in two places. Please see my comment on 'accept interfaces return structs'. I would mark this thread as solved. In general I'm fine with the PR

EnricoSchw: To your example:
I think a caller with a hash-map of subscriptions should knows (define by him self) the structure about what the he save. 
The publisher only have to know thats a subscriptions has a method: `WriteRTP(packet rtp.Packet) error` thats it.
The Caller should not ask the Publisher what he knows about subscriptions?

Maybe the name of this subscription interface is miss leading. In idiomatic Go we would name the interface like:

```Go
type rtpWriter interface {
	// WriteRTP **must not** block (wait on I/O).
	WriteRTP(packet rtp.Packet) error
}
```

EnricoSchw: Maybe another point to this view. (Even if I'm starting a fundamental discussion here 👀 )

Patterns like Factory do not makes sense in Go.

daniel-abramov: > The second point was, how we deal with signal channels from parent go routines.
> A child should tell parent routine about close than reaction on parent routine closing.

Do you mean that the parent should never inform the child that they're done, signaling the child to stop?

> I think a caller with a hash-map of subscriptions should knows (define by him self) the structure about what the he save.

That is true, yet the caller may (for whatever reason) store the list of all available subscriptions in a collection, let's say in a map or in an array. The map or an array can be a collection of heterogenous types, all of which only have one thing in common: they are all subscriptions. This way, the caller may store them `map[string]publisher.Subscription`, otherwise there will be no other option apart from `map[string]interface{}` to store heterogeneous types in a single collection.

We as publishers don't care about the exact type or the implementation details of the subscriptions, we're ok accepting any type as long as it satisfies the contract (the interface), the actual subscription could be anything from a WebRTC track to just plain RTP sink or any other stream that accepts rtp packets 🙂 

> Patterns like Factory do not makes sense in Go.

Sorry, I did not get what you meant by it (in particular, what it related to 🙂).

Issue #133, opened 2023-02-14
To improve session performance: move blocking operations out of the main conference loop


daniel-abramov:
It looks like sending a message on a data channel is a blocking operation (due to the I/O involved in it), in my non-scientific tests it turned out that sending a message on a data channel may sometimes take 30ms. That's not a problem in the majority of cases, but when the conference gets large and participants to exchange messages over the data channel, it can delay the processing of the incoming To-Device messages significantly.

daniel-abramov:
I tried to measure the amount of time it takes to process each conference message in the main loop to see which of them takes most of the most time. In a conference with 4 participants, the majority of the messages take less than 1ms to be handled.

However, the same message may under certain circumstances take more time to get processed.

I tried to log all occurrences when a message took more than 1ms to handle and sorted them in a descending order:

```
onNewParticipant:                            17.938334ms
onNewParticipant:                            14.249709ms
conference.processNewTrackPublishedMessage:  11.006ms
resendMetadataToAllExcept:                   10.72575ms
onNewParticipant:                            9.574333ms
conference.processNewTrackPublishedMessage:  8.04675ms
resendMetadataToAllExcept:                   7.9625ms
conference.processDataChannelMessage:        7.410416ms
onNewParticipant:                            7.355125ms
conference.processNewTrackPublishedMessage:  6.806458ms
conference.processDataChannelMessage:        5.301042ms
resendMetadataToAllExcept:                   4.100334ms
conference.processLeftTheCallMessage:        4.078875ms
removeParticipant:                           4.051708ms
conference.processNewTrackPublishedMessage:  3.850084ms
resendMetadataToAllExcept:                   3.667958ms
onNewParticipant:                            3.462417ms
onHangup:                                    3.393917ms
removeParticipant:                           3.294875ms
onNewParticipant:                            2.74225ms
onCandidates:                                2.324042ms
conference.processDataChannelMessage:        2.282125ms
conference.processNewTrackPublishedMessage:  2.213209ms
getAvailableStreamsFor:                      2.178167ms
onCandidates:                                2.138333ms
conference.processDataChannelMessage:        2.096166ms
resendMetadataToAllExcept:                   2.037042ms
onNewParticipant:                            1.987ms
onNewParticipant:                            1.858708ms
conference.processDataChannelMessage:        1.856125ms
conference.processTrackSubscriptionMessage:  1.651292ms
conference.processDataChannelMessage:        1.609167ms
[...]

PR #136, opened 2023-02-23
peer: add data channel worker

daniel-abramov:
This relates to https://github.com/matrix-org/waterfall/issues/133

The effect of this change is not as large as anticipated though. It turns out that for smaller messages in conferences with up to 6 people, the operation of sending to the data channel on the peer level took no more than 20µs on average. The peaks occur from time to time (up to 1 ms), but they are not as large, so the effect of this worker is not really that noticeable.

So now I question if it's worth it at all. Decided to file a PR to at least demonstrate how simple and small the change was and where it was done. We can close if we think that the complexity does not justify the improvement.

Please check the details: https://github.com/matrix-org/waterfall/issues/133#issuecomment-1441729488

### DIFF
--- a/pkg/peer/datachannel.go
+++ b/pkg/peer/datachannel.go
@@ -1,0 +1,33 @@
+package peer
+
+import (
+	"time"
+
+	"github.com/matrix-org/waterfall/pkg/peer/state"
+	"github.com/matrix-org/waterfall/pkg/worker"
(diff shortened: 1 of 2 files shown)


dbkr (approved):
Looks good: we can always try it & see if it makes a difference and back it out if it doesn't seem worthwhile.

Issue #121, opened 2023-02-01
Find out why we have so many warnings on missing RTPs on certain active subscriptions


daniel-abramov:
A log line like this one is not uncommon during normal sessions:

```
waterfall  | WARN[2023-02-01T18:19:37Z] No RTP on subscription 9db59e08-6390-44d5-a77a-11ced35eee2d (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=RWSECBPMWU user_id="@copper-sunny-butterfly:call.ems.host"
waterfall  | WARN[2023-02-01T18:19:38Z] No RTP on subscription 9db59e08-6390-44d5-a77a-11ced35eee2d (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=AZGGKGVAFT user_id="@robintown:call.ems.host"
waterfall  | WARN[2023-02-01T18:19:38Z] No RTP on subscription 6df7aea5-1bfa-45c6-91a6-87422782a541 (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=AZGGKGVAFT user_id="@robintown:call.ems.host"
waterfall  | WARN[2023-02-01T18:19:38Z] No RTP on subscription 9db59e08-6390-44d5-a77a-11ced35eee2d (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=YSMNDCBZWH user_id="@copper-resident-stingray:call.ems.host"
waterfall  | WARN[2023-02-01T18:19:38Z] No RTP on subscription 9db59e08-6390-44d5-a77a-11ced35eee2d (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=WDEKNIAADZ user_id="@cyan-patient-junglefowl:call.ems.host"
waterfall  | WARN[2023-02-01T18:19:38Z] No RTP on subscription 9db59e08-6390-44d5-a77a-11ced35eee2d (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=NUOTIRTNCN user_id="@simon.brandner:call.ems.host"
waterfall  | WARN[2023-02-01T18:19:38Z] No RTP on subscription 6df7aea5-1bfa-45c6-91a6-87422782a541 (low)  conf_id=1675175890574TP04z8LAccuowlpZ device_id=YSMNDCBZWH user_id="@copper-resident-stingray:call.ems.host"
[...]

PR #137, opened 2023-02-23
subscription: change the warning message

daniel-abramov:
This message seems to happen primarily when the video is muted, but seldomly in other cases that are yet to be discovered.

Printint the warning in the logs does not make any sense as our logs end up being filled up with such messages. One could implement a reaction to this message in a way that we'll stop the given subscription if it's stalled (whether for mute reason [video]) or for any other reason.

But this means that the SFU will actively unsubscribe the client from the subscriptions that appear dead for the SFU. The question is if such behavior is something that the EC is ready for? Since currently we don't have a mechanism to inform the EC for the changes (well, we could remove the track that will possibly cause a renegotiation, but that's it).

@SimonBrandner do you know how the EC would behave if one of the subscriptions gets closed by the server? Will the EC try to re-request it if necessary? (if so, I could implement the logic on the SFU side to remove stalled subscriptions and remove the corresponding tracks from the peer connection that they are related to)

Relates to https://github.com/matrix-org/waterfall/issues/121

### DIFF
--- a/pkg/conference/subscription/video.go
+++ b/pkg/conference/subscription/video.go
@@ -71,14 +71,13 @@ func NewVideoSubscription(
 
 	// Configure the worker for the subscription.
 	workerConfig := worker.Config[rtp.Packet]{
-		ChannelSize: 16,              // We really don't need a large buffer here, just to account for spikes.
-		Timeout:     3 * time.Second, // When do we assume the subscription is stalled.
+		ChannelSize: 16,               // We really don't need a large buffer here, just to account for spikes.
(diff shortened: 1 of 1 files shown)


dbkr (approved):

SimonBrandner:
> @SimonBrandner do you know how the EC would behave if one of the subscriptions gets closed by the server? Will the EC try to re-request it if necessary? (if so, I could implement the logic on the SFU side to remove stalled subscriptions and remove the corresponding tracks from the peer connection that they are related to)

Yeah,it would end-up requesting it eventually due to window resizing etc. but there is no clear logic for it atm

Issue #100, opened 2023-01-12
Make switching between layers seamless (without any freezes or artifacts)


daniel-abramov:
Sometimes when we switch between layers, there is a freeze of up to 2 seconds. Sometimes instead of a frozen picture, one can observe a short glitch for half a second. This happens because when we switch between the layers, we must get a keyframe to ensure smooth playback. We do request a new keyframe, but it may take some time until we get one (2 RTT + processing).

We do not store the previous keyframes (and should not), however, what we could do is implement a logic where when we switch the layer, we continue to send frames on a previously selected layer (as if the layer was not switched) up until we get the new keyframe on a layer that we wanted to switch to. Once the keyframe is received, we switch to the newest layer.

That way there won't be any glitches observed during the switch.

PR #140, opened 2023-03-06
rewriter: add vp8 keyframe detection function

daniel-abramov:
We've postponed working on https://github.com/matrix-org/waterfall/issues/100

The branch I started more than a month ago is obsolete and does not compile anymore (due to conflicts caused by changes done in the previous PRs). Still, the function for the detection of VP8 frames is somewhat isolated and does not really have any dependencies, so I decided to file the PR with this small function so that I can remove the now obsolete branch while preserving something that could be reused by one who works on the issue in the future.

### DIFF
--- a/pkg/conference/subscription/rewriter/vp8.go
+++ b/pkg/conference/subscription/rewriter/vp8.go
@@ -1,0 +1,29 @@
+package rewriter
+
+import (
+	"github.com/pion/rtp"
+	"github.com/pion/rtp/codecs"
+)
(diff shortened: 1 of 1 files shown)


SimonBrandner (approved):
I am not sure how big of a fan of dead code I am but this seems okayish

daniel-abramov:
> I am not sure how big of a fan of dead code I am but this seems okayish

I'm with you on that 🙂

PR #144, opened 2023-03-21
Refine error logs, primarily the data channel logs

daniel-abramov:


pkg/conference/publisher/publisher.go
```diff
@@ -54,7 +54,7 @@ func NewPublisher(
 				return
 			default:
 				if err := publisher.forwardPacket(); err != nil {
-					log.Errorf("track ended: %s", err)
+					log.Infof("track ended: %s", err)
```
EnricoSchw, line 57: This is not an error anymore?

EnricoSchw, line 57:
```suggestion
					log.Infof("Forwarding packages for track ended: %v", err)
```

daniel-abramov: I've noticed that this error has been logged all the time when the tracks stopped gracefully, i.e. with the EOF, so it was not actually an error, but a normal lifecycle of a track.

But now, after your comment, I realize that I probably must not treat **each error** returned from the `forwardPacket()` as "graceful close", instead I must check if it was EOF (normal close) or if something abnormal happened and log them differently. I'll address it right away, thanks for pointing it out!

pkg/conference/publisher/publisher.go
```diff
@@ -54,7 +55,11 @@ func NewPublisher(
 				return
 			default:
 				if err := publisher.forwardPacket(); err != nil {
-					log.Errorf("track ended: %s", err)
+					if err != io.EOF {
+						log.Errorf("Track failed (%s)", err)
```
EnricoSchw, line 59:
```suggestion
						log.Errorf("Forwarding packages for track failed: %v", err)

```
I would not log the error as string. And i would log what we tried to do here, because this fails.  The previous reason (something with the track fails), should be inside the `err` message.

daniel-abramov: > I would not log the error as string.

Are not `%v` and `%s` identical for **errors**? `%s` would print the string representation, whereas `%v` the value in default format, we arguably need `%s` for this (although in case of error they would be identical, right?).

> And i would log what we tried to do here, because this fails.

But will it be as useful (without the additional knowledge of what it means), i.e. "Forwarding packages for track failed" may be interpreted as "certain packets were not forwarded", but does not necessarily mean that the track has "failed" or unusable anymore, whereas with "Track failed" we signify that the track has valid and is invalid anymore and won't be used (which is exactly what would happen in this case)?

A small thing, but still.

EnricoSchw: > Maybe you meant the `%w`?

I assumed you couldn't use `%w`. If you can use `%w` then, please use this. If you can not use `%w`, please use `%v` . The behavior of `%v` is better than `%s` in some cases.

> But will it be as useful... 

Why is is this so useful? 

The track failed not here on this place.  The track failed inside `publisher.forwardPacket()` And no even there it was inside  `track.ReadPacket()` And no even there it was...  And so on!

Would we logging what we doing like this in side `publisher.forwardPacket()`:

```go
packet, err := track.ReadPacket()
	if err != nil {
		return fmt.Errorf("reading track packages:  %w" , err)
	}
```	

we would get an error log output like this:

`forwarding Packages for publisher: reading track packages: can't read track: ....`

And this would really useful in production. Because as Admin without knowing the source code We would see and understand what happen.

To get something like this and Because there is no extra log formatting inside `publisher.forwardPacket()` I suggested:  `Forwarding packages for track failed: %v",`

daniel-abramov: Sorry @EnricoSchw, I think there might be a misunderstanding, let me elaborate.

The purpose of the log line in question is not to inform about the error while forwarding packets, but rather to log that **the publisher has been stopped**. See, the context of this particular part of the code is that it's the "main loop" of the "publisher thread" (publisher go-routine). Currently, we assume that errors from internal functions (`forwardPacket()`) are fatal and so we end the go-routine (see `return`).

So the purpose of the logline is to barely log the fact that the publisher has been stopped while providing the information of why the stop occurred.

The reason I was against using the formulation "Forwarding packets failed" or "Forward packets stopped" is because it's just a complicated way of saying that the publisher has stopped, i.e. as an abstract administrator who observes log (the one that you mentioned in your comment 🙂), you don't know what "Forwarding packets stopped" means. So instead of logging what happened (a function failed, this is the implementation detail), I logged **what it actually means**, i.e. the publisher has stopped. Perhaps my mistake was using the wording "track stopped" instead of "publisher stopped" (i.e. assuming that a publisher is associated with a track that it represents). I've changed it to the more neutral "publisher", so now it matches the corresponding formulating we had for the audio publisher. Does it make any sense? 🙂

As for `%s` vs `%v` vs `%w`. The thing is:
- We could use `%w` for errors (they support it), but it does not look nice when printed (so it's more suitable when returning errors I think).
- `%v` and `%s` are identical when logging errors, except when the error is `nil` (or the `Format()` which is unusual from the errors that we handle). The error cannot be `nil` since we only execute the code if the condition `if forwardPacket() != nil` 😉, which means that `%s` and `%v` would look identical when logged. I've seen both flavors of handling errors in the Go code. I changed it to `%v` as you suggested since I don't really have a strong preference.

Ptal, I hope this clarifies things a bit 🙂

EnricoSchw (approved):

PR #146, opened 2023-03-24
subscription: don't assume muted tracks stalled

daniel-abramov:
When the user mutes the video, we don't get any RTPs. We don't assume this error anymore and don't change the status of the track and its telemetry for that.

This removes false positives related to the
https://github.com/matrix-org/waterfall/issues/121

### DIFF
--- a/pkg/conference/participant/tracker.go
+++ b/pkg/conference/participant/tracker.go
@@ -161,7 +161,11 @@ func (t *Tracker) RemovePublishedTrack(id track.TrackID) {
 }
 
 // Subscribes a given participant to the track.
-func (t *Tracker) Subscribe(participantID ID, trackID track.TrackID, requirements track.TrackMetadata) error {
+func (t *Tracker) Subscribe(
+	participantID ID,
+	trackID track.TrackID,
+	desiredWidth, desiredHeight int,
+) error {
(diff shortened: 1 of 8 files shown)


pkg/conference/participant/tracker.go
```diff
@@ -161,7 +161,11 @@ func (t *Tracker) RemovePublishedTrack(id track.TrackID) {
 }
 
 // Subscribes a given participant to the track.
-func (t *Tracker) Subscribe(participantID ID, trackID track.TrackID, requirements track.TrackMetadata) error {
+func (t *Tracker) Subscribe(
+	participantID ID,
+	trackID track.TrackID,
+	desiredWidth, desiredHeight int,
+) error {
```
dbkr, line 168: This change doesn't seem related?

dbkr (approved):
Otherwise lgtm

pkg/conference/participant/tracker.go
daniel-abramov: I changed the argument from `TrackMetadata` to `desiredWidth, desiredHeight`. Previously these were the only fields in the `TrackMetadata`. Now, it's more extensive (contains the mute state and probably will contain more), i.e. it was required since I changed the input parameters for the `published.Subscribe()` function.

PR #147, opened 2023-04-11
Detect stalled publishers (layers) and switch to other publishers (layers)

daniel-abramov:
When simulcast is used, we get different publishers (a single publisher per simulcast track/layer). The browser may decide to stop sending RTPs to specific publishers, so we will stop getting RTP packets.

If there is a subscriber X that is subscribed to publisher Y and publisher Y does not emit any frames, subscriber X will observer it as a "video freeze" (if it's in the middle of the conference) or as a "grey frame" (if it's at the beginning of the conference).

We solve the issue by detecting such stalled publishers Y. Then we gather the set of subscribers that are subscribed to the given publisher and let them receive RTP frames from a different publisher Z (as long as the publisher belongs to the same track), i.e. effectively we detect when e.g. 3 users are subscribed to the high-quality video of a user Y and then we notice that user Y stops sending the high quality, but continues to send the low quality, we switch all these subscribers to the low quality.


pkg/conference/track/track.go
```diff
@@ -130,6 +114,15 @@ func NewPublishedTrack[SubscriberID SubscriberIdentifier](
 		defer close(published.done)
 		defer telemetry.End()
 		published.activePublishers.Wait()
+
+		// End any active subscriptions.
+		published.mutex.Lock()
+		published.mutex.Unlock()
```
dbkr, line 120: defer?

dbkr (commented):
Otherwise looks good!

pkg/conference/track/track.go
daniel-abramov: Wow! Thanks! I really missed it, glad that you've found it!

Debugging it once it leads to a race would def took a lot of time 🙂

Fixed.

dbkr (approved):

PR #149, opened 2023-04-21
track: fix incorrect subscription existence check

daniel-abramov:
If someone tries to subscribe to an audio or non-simulcast track and the subscription exists already, we must not create a new subscription.

The implementation in `main` had a bug where each client entered a vicious cycle of adding a subscription -> getting the renegotiation from Pion -> sending it to the remote counterpart of the client -> adding a new track -> subscribing again -> again getting the renegotiation from Pion -> ... etc.

The only thing that is not quite clear is why the client tried to subscribe to the same track many times. Regardless of this, the SFU should not have created new subscriptions for the subscriptions that already exist, this PR fixes that.

### DIFF
--- a/pkg/conference/track/track.go
+++ b/pkg/conference/track/track.go
@@ -187,33 +187,38 @@ func (p *PublishedTrack[SubscriberID]) Subscribe(
 	p.mutex.Lock()
 	defer p.mutex.Unlock()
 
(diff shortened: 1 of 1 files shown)


SimonBrandner (approved):
Looks sane

